PR #13, opened 2025-07-12
`..()` to `. = ..()`

FalloutFalcon:
## About The Pull Request
Most procs do acctually care about there return values (primarly initiziate) but flav is a bit dumb and used `..()` in alot of places meaning code wont return its parents return value which is bad, I've left out it `New` and alot of outfit code as both dont care about return values in tg
this is also restricted to our modular folder exlusivly


TheCarnalest (approved):
Good change, thanks Fallcon!!

PR #14, opened 2025-07-12
bank accont and credit card repathing and minor refactors

FalloutFalcon:
## About The Pull Request
removes the vtm_bank_account type cause its dum and we should just use bank_account 
IDs have had all account related behavoir removed as im not buying chips from the gas station using my hospital id thats silly (it still has the definition cause i got tired of fixing stuff for content that we are never gonna use. will do a further cleanup of that later)
Code for having wallets and debit cards defining in job types instead of being added to backpack contents (not implmented since wallets  are undercooked and debit cards still exist in most backpack contents

- [x] Remove bank stuff from IDs (but more modular this time?)

## Why It's Good For The Game
less bad code. sets up code to make it easier to use debit cards for cash based interactions.

## Changelog

:cl:
del: IDs no longer have bank accounts attached to them.
qol: Debit cards have replaced IDs for most cash based interactions
/:cl:


modular_darkpack/modules/economy/code/credit_card.dm
```diff
@@ -1,12 +1,122 @@
...
+				account.bank_cards += card
+
+
+/obj/item/proc/GetCreditCard()
+	return null
+
+/obj/item/storage/wallet/GetCreditCard()
+	for(var/obj/item/card/I in contents)
+		if(is_creditcard(I))
+			return I
```
XeonMations, line 99:
```suggestion
	var/obj/item/card/I = locate(//obj/item/card/) in contents
	return I
```

modular_darkpack/modules/werewolf_the_apocalypse/code/garou/garou_species.dm
```diff
@@ -54,10 +54,12 @@
 			if(!host.mind.assigned_role)
 				dat += "."
 			dat += "</p>"
-		for(var/datum/vtm_bank_account/account in GLOB.bank_account_list)
-			if(host.bank_id == account.bank_id)
-				dat += "<p>My bank account code is: [account.code]</b></p>"
-				break
+		if(istype(host, /mob/living/carbon/human))
```
XeonMations, line 57: ``host`` is already assumed to be ``/mob/living/carbon/human`` at this point

XeonMations (changes requested):

modular_darkpack/modules/economy/code/credit_card.dm
FalloutFalcon: should actually be equivalent to 
```c++
/obj/item/storage/wallet/GetCreditCard()
	return locate(/obj/item/card/credit) in contents
``` 
as well

modular_darkpack/modules/werewolf_the_apocalypse/code/garou/garou_species.dm
FalloutFalcon: fixed!

modular_darkpack/modules/economy/readme.md
```diff
@@ -0,0 +1,46 @@
...
+
+Module ID: ECONOMY <!-- Uppercase, UNDERSCORE_CONNECTED name of your module, that you use to mark files. This is so people can case-sensitive search for your edits, if any. -->
+
+### Description:
+
+<!-- Here, try to describe what your PR does, what features it provides and any other directly useful information. -->
+
+### TG Proc/File Changes:
+
+- N/A
```
TheCarnalest, line 15: EXTREMELY MINOR, but mind removing this?

TheCarnalest (approved):

TheCarnalest (approved):

TheCarnalest:
hmm, failed unit test, that's odd?

modular_darkpack/master_files/code/modules/economy/account.dm
```diff
@@ -0,0 +1,14 @@
...
+		bank_code += "[rand(0, 9)]"
+	return bank_code
+
+/datum/bank_account
+	var/bank_pin
+
+/datum/bank_account/New(newname, job, modifier = 1, player_account = TRUE)
+	. = ..()
+	if(!bank_pin)
+		bank_pin = create_bank_pin()
```
XeonMations, line 14: Surely there's an easier way of generating a bank pin that doesn't involve the usage of a global proc and a for loop?

modular_darkpack/modules/economy/code/credit_card.dm
```diff
@@ -1,12 +1,120 @@
...
+	if(!.) //If there is no id, check the other hand
+		held_item = get_inactive_held_item()
+		if(held_item)
+			. = held_item.GetCreditCard()
+
+/mob/living/proc/get_creditcard_in_hand()
+	var/obj/item/held_item = get_active_held_item()
+	if(!held_item)
+		return
+	return held_item.GetCreditCard()
```
XeonMations, line 119: What's the point of all of those procs when you can just do ``locate() in mob``?

XeonMations: Not to mention we probably want to only be able to tell if a card is existent if it's inside of a machine, like real life.

XeonMations (changes requested):

modular_darkpack/modules/economy/code/credit_card.dm
FalloutFalcon: this purposefully only searching your hands and your wallet. Being able to spend with a credit card from inside your backpack is prob weird.

modular_darkpack/master_files/code/modules/economy/account.dm
FalloutFalcon: uhh its like that because it wants to be a string, if you just did rand(0,9999) you would get like, 1 rather then 0001|
I could make it bank_code = "[rand(0, 9)][rand(0, 9)][rand(0, 9)][rand(0, 9)]" which, is prob more efficent?
and the fact its a global proc is kinda whatever. just incase something besides bank accounts want to make up a fake pin

XeonMations: we wont have anything that requires a fake pin in the near future. If we do, then we can make it an appropriate proc. Creating unnecessary global procs is needlessly clogging up VSC's procs list currently.

modular_darkpack/modules/economy/code/credit_card.dm
XeonMations: Why not just have the player be required to insert their card or swipe it to spend it, and grab the associated bank account that way?

FalloutFalcon: because thats how most tg code grabs ids right now. I can do a pr later for inserting/swiping ids in some machines but we want this for cross compatability with tg machines.

XeonMations (approved):

PR #28, opened 2025-07-13
Implement aggravated damage

TheCarnalest:
## About The Pull Request
Modern TG has done away with CLONE damage since they no longer have cloning as a feature. This is a problem because the old WoD codebase made heavy use of CLONE damage as a stand-in for aggravated damage as it appears in tabletop. This PR implements a new AGGRAVATED damage type for supernatural damage so that we can give it special handling, damage mechanics, and healing mechanics.

This adds full compatibility for dealing and healing aggravated damage, along with adding it to relevant places like examine and a component we use for Disciplines. 

This does not add aggravated damage dealing and healing (or even reading, in the case of medical scanners) to every place it'd make sense. I'd rather not bloat the PR more, I'll just leave that for when those other bits of code get used.

## Why It's Good For The Game
If we don't want to have our aggravated damage type be BURN, this is a major project we need to undergo as part of the rebase.

## Changelog

:cl:
refactor: Aggravated damage no longer uses clone damage, and instead has its own unique damage type.
/:cl:


code/modules/surgery/bodyparts/_bodyparts.dm
```diff
@@ -296,6 +297,10 @@
 		. += span_warning("This limb has [brute_dam > 30 ? "severe" : "minor"] bruising.")
 	if(burn_dam > DAMAGE_PRECISION)
 		. += span_warning("This limb has [burn_dam > 30 ? "severe" : "minor"] burns.")
+	// DARKPACK EDIT ADDITION START - AGGRAVATED_DAMAGE
+	if(aggravated_dam > DAMAGE_PRECISION)
+		. += span_warning("This limb has [aggravated_dam > 30 ? "severe" : "minor"] cellular damage.")
```
FalloutFalcon, line 302: cellular?

FalloutFalcon (approved):
Everything else looks good. Definitely the type of pr that SHOULD get a test merge if the repo was in a more complete state. It might need more fleshing out once we have playtesters but all the bones are here.

code/modules/surgery/bodyparts/_bodyparts.dm
TheCarnalest: I have no clue what to call it flavour wise, but I replaced it with festering wounds now

PR #66, opened 2025-08-21
exludes stations from ci and map voting

FalloutFalcon:
## About The Pull Request
Removes station configs so they wont be votable or run tests for them.
Not aware of a better implementation for this right now.

## Why It's Good For The Game
Our tests have been insanely flakey for like tramstation specificly but we also dont need to test any base ss13 maps other then mabye the debug ones both now and when we have real maps.


FalloutFalcon:
> why not just comment out the maps.txt and leave the json as is

for the second part, linters should compile a list based of the jsons and entirely ignore the txts

PR #103, opened 2025-09-24
Darkpack Cats

KCartridge:
## About The Pull Request

Re-implements the stray cats found in da city.
Also adds a rest sprite with a lil wagging tail that I made :) 

Plans are at some point down the road I'll add held sprites and sitting sprites for strays, I just wanted to get the PR out.
This PR also makes them basicmobs as opposed to simplemobs, which might be fun.

## Why It's Good For The Game

Anotha thing ticked off the list YIPPEE

## Changelog

:cl: KCartridge
add: Stray Cats!
image: cats can now hypothetically rest, and their lil tail wags when they do
code: cats are now basic instead of simple mobs
/:cl:


FalloutFalcon (dismissed):

code/modules/mob/living/basic/pets/cat/cat.dm
XeonMations, line 1: Wouldn't it be better if these were just a subtype of the default cats in its own modular file instead of adding edits like this?

XeonMations (changes requested):

code/modules/mob/living/basic/pets/cat/cat.dm
KCartridge: idk maybe

PR #119, opened 2025-10-04
Wallpaper file org and minor ports

FalloutFalcon:
## About The Pull Request

ports some assets that are in tfn (padded wallpaper and wallpaper**s**)
also wallpapers are no longer on CAR layer lol lmao
breaks up wallpaper into there own dmi and dm
makes sure you can place wallpapers on walls (tho they should prob have directionals?)


XeonMations:
IMO wallpapers should just be wall subtypes tbh

FalloutFalcon:
> might want to add a linter test to make sure wallpaper is always placed on a wall instead of the air

the fucked thing is BOTH is valid and used in mapping right now.

PR #98, opened 2025-09-22
Improve Jumping Component; separate it from pointing code

hex37:
- Prevents the click from Jumping from interfering with control/alt/shift clicking
- The do_mob bar on jumps can no longer be spammed

PR #75, opened 2025-08-28
Adds jumping

hex37:
## About The Pull Request
- Readds code for jumping, packaged into a component that can be added to any living mob.
- Jumping now has a windup time based on the user's athletics.
- Jumping consumes stamina on use.
- Ground-Pounding requires 8+ strength and combat-intent to perform, and consumes additional stamina.
- Adds and simplifies the UI Element for the jump code
- Shuffles around some blood drinking icon paths so they are no longer in depreciated files.

## Why It's Good For The Game
Jump Good.

## Changelog
:cl:
add: Jumping Component
add: Jumping UI Button, for when we do the interface
refactor: Component is added to all human mobs.
del: Old Jump Code
/:cl:

PR #146, opened 2025-10-11
Post-Jump slowdown replaces windup (with skill checks)

dwinters99:
## About The Pull Request
Puts jump windup on a config option, replacing it with post-jump slowdown.

I've addressed balance concerns leveled in [this conversation on the discord.](https://ptb.discord.com/channels/1393132000583618701/1396599047128617010/1423933973729509379)

When two characters collide, both the jumper and the victim make a contested roll. If the victim succeeds, they do not get knocked down or disarmed.

The winner is calculated thusly:
We check the jumper's Strength + Brawl and Dex + Athletics and keep the higher of the two values. We then check the victim's Stamina + Athletics or Stamina + Brawl, whichever's higher. Finally, we roll these values against a median difficulty of 6 (as per medium difficulty in V20 and the standard V5 rolling rules). The owner of the higher value wins. In cases of a tie, the jumper wins. The loser is paralyzed, knocking them down and disarming them.

Regardless of whether or not the jumper impacts something during the jump, they are slowed in such a way that it would have been approx. 1 tile faster to have walked that distance assuming they wish to continue moving in that direction. This both punishes jumping as a casual traversal method and provides a window for agile opponents to punish the jumper if they are jumping recklessly.

The latter change was implemented on Apocrypha XIII in https://github.com/ApocryphaXIII/ApocryphaXIII/pull/183 and has had extensive testing, of which feedback was overwhelmingly positive. This change will also be tied to a config option.

Thank you to @hex37 for discussing these balance concerns with me and originally implementing jumping to SecondCity in #75 and #98.

Depends on #144 

https://github.com/user-attachments/assets/91fc443f-d5ff-4d95-8a09-503c14486bb5

## Why It's Good For The Game
Players on Apocrypha XIII and a few members of The Final Nights have come to me variously regarding the [aforementioned PR](https://github.com/ApocryphaXIII/ApocryphaXIII/pull/183) with extremely strong opinions AGAINST jump windup. That being said, Requiem still seems to desire jump windup so it has been left as a config option if it is desired.

## Changelog

:cl:
balance: Replaced jump windup with post-jump slowdown
code: Added skillchecks to carbon throw_impact code
config: Jump windup and jump slowdown config options
/:cl:


FalloutFalcon:
conflicts! @dwinters99

FalloutFalcon (approved):
otherwise good!

PR #160, opened 2025-10-14
Diceroll difficulty defines

dwinters99:
## About The Pull Request
Difficulty defines corresponding with p249 of the V20 corebook
## Why It's Good For The Game
Easier to spitball difficulties for rolls (i.e. knock that up from easy to standard)
## Changelog
:cl:
code: Roll difficulty defines
/:cl:


XeonMations:
> > personally I dont see any reason for this
> 
> It makes it easier to CTRL-F for static difficulties rather than having to sort by, like `, 6,` which is very likely to be an unhelpful search. This is a spite PR caused by me trying to find storyteller rolls with static difficulties.

why not just do like ``DIFFUCLTY_6`` then

chazzyjazzy:
> > > personally I dont see any reason for this
> > 
> > 
> > It makes it easier to CTRL-F for static difficulties rather than having to sort by, like `, 6,` which is very likely to be an unhelpful search. This is a spite PR caused by me trying to find storyteller rolls with static difficulties.
> 
> why not just do like `DIFFICULTY_6` then

+1 to this, if im searching for static defines i would have a harder time looking for difficulty extreme than i would with difficulty_8 or whatever

PR #222, opened 2025-11-04
Fixes a lot of missing items for maps

XeonMations:
## Changelog
:cl:
map: Updated updatepaths scripts to not miss some stuff.
/:cl:


modular_darkpack/modules/piano/code/stationary_pianos.dm
FalloutFalcon: im not sure why ipiano exists, it seems to be a diretional for the normal piano sprite
would much rather just give that diretionals.
the darkpack piano type also seems unessicary. would just overide the name of the tg one to remove the spess refrences

PR #266, opened 2025-11-15
Normal people cant craft tzi stuff and arcane books

FalloutFalcon:
## About The Pull Request
Lock these behind the disciplines they are meant to require
## Changelog
:cl:
fix: Tzi content and arcane book must be learned
/:cl:


XeonMations (changes requested):
Tzimisce crating recipes are already locked behind viscissitude, but the code is commented out since we dont have the discipline in yet. The books should probably be done the same.

FalloutFalcon:
> Tzimisce crating recipes are already locked behind viscissitude, but the code is commented out since we dont have the discipline in yet. The books should probably be done the same.

I went and made it not ass

PR #332, opened 2025-11-27
Adding a Ventrue Law Office, Groundskeeping Area, Art store/FBI office, and Candy Store

Drakeven:
## About The Pull Request
Filled out most all the lower right area with things. Might need to do some further tweaking with the Ventrue area, but I will do that in future commits.

Here's a vague image: 
image
image

## Why It's Good For The Game
More things!

## Changelog
:cl:
map: More stuff to the westfield mall map
/:cl:

### DIFF
--- a/_maps/map_files/Vampire/westfield_mall.dmm
+++ b/_maps/map_files/Vampire/westfield_mall.dmm
@@ -16998,7 +16998,7 @@
 /area/vtm/interior/shop)
 "mFb" = (
 /obj/machinery/door/poddoor/shutters{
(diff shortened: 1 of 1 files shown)


FalloutFalcon:
this is projectile and i donnt think can be realsticly used in the map. swap it to `/obj/item/ammo_casing/rebar/paperball` which has the same sprite.
image

rapier sheathe comes preloaded, dont need the extra sword
image

Color pallet seems a bit.. messy here.. 
image

this is the 911 phone..
image

PR #411, opened 2025-11-30
ports derangement

buffyuwu:
## About The Pull Request

ports malkavian derangement quirk, including shifting tiles and objects speaking. also refactors malk.txt into a new global_strings.dm file where we can put our big annoying glob lists that get used in more than one place. example: the associative list used for the hallucination messages and sounds in this PR will be used in the dementation ability

based on the work here: https://github.com/The-Final-Nights/The-Final-Nights/pull/287

put the sounds under the dementation discipline as they will likely be used by it as often as this quirk uses them. adds the whispers sounds malkavians experience from the original bloodlines game:

https://github.com/user-attachments/assets/af98e530-a0db-40fc-bf50-15d2f53d2178

if you want me to reorganize any of these files lmk

## Why It's Good For The Game

malkavian features good
dreamseeker_5fTdeE6gOZ

## Changelog
:cl:
add: ports derangement quirk, basic malk hallucinations from bloodlines
/:cl:


buffyuwu:
> looks good but fix yo linters

im still cranking my hog trying to get it to copy over the appearance, but linters look good

buffyuwu:
> I wrote up this whole thing on why derangement shouldn't be a quirk, forgetting it is actually just a flaw normally. I actually am > changing this to say its weird that the current derangement quirk is written elusive to malks when we should have a generic flaw > it seems??? I haven't read too hard, consult with others.

i talked to xeon and we agree its probably best served as a trait so ill convert it into one. if you decide to keep it a quirk i can just roll back the trait commit :)

> Can you verify that it doesn't show up as select-able in quirks right now as well?
> 
> Skimming Apoc it also semes like your derangment used to be chooseable.

yeah in the current implementation theres no actual quirk blacklist, only the ability to choose traits that are incompatible with one another. we'd have to implement additional quirk overrides if we keep it a quirk but want it to be, say, malk only or whatnot

> This is missing some stuff I think, there is a buncha non-modular changes that was made to the base hallucinations to make them more Wod/vtm adjacent.

there's some extremely chopped vtm13 hallucination code missing from this PR that is present on tfn yeah, i'd like to have the more advanced malk hallucinations (outside of the talking objects in this PR) be another project altogether rather than continue to carry the stinky code into the clean new codebase

chazzyjazzy:
> i talked to xeon and we agree its probably best served as a trait so ill convert it into one. if you decide to keep it a quirk i can just roll back the trait commit :)


so what will malks get then that is unique to their clan curse?

chazzyjazzy:
> > > i talked to xeon and we agree its probably best served as a trait so ill convert it into one. if you decide to keep it a quirk i can just roll back the trait commit :)
> > 
> > 
> > so what will malks get then that is unique to their clan curse?
> 
> From what I can tell from, the book, there is nothing malk specific to the derangement, its just that they have an unremovable one. image

so any kindred will be able to have malkavian derangements but malkavians are the ones who are uniquely stuck with it? similar to the unbonding trait and the tremere v5 curse? 

speaking of perhaps we should swap the tremere v5 curse for the v20 curse if we rework bloodbonds. maybe i should work on that but savefile stuff is so eugh

chazzyjazzy (approved):
dont really see any other way of implementing this

XeonMations (changes requested):
lowertext your fucking filenames, especially your sound files

chazzyjazzy:
ill be working on this for Nimi

chazzyjazzy (approved):

PR #434, opened 2025-12-12
Reimplements ATMs

FalloutFalcon:
## About The Pull Request
Refactors a bit of the ATM code and gives it a decent theme. 
image
## Changelog
:cl:
add: Bank pin is now found in the memories section similar to bank ID number
/:cl:


XeonMations (changes requested):

modular_darkpack/modules/economy/code/atm.dm
```diff
@@ -1,152 +1,148 @@
...
+	while(amount > 0)
+		var/drop_amount = min(amount, 1000)
+		if(drop_amount > total_stored_cash)
+			to_chat(user, span_notice("[src] has maxed out its withdraw limit"))
+			break
+		var/obj/item/stack/dollar/cash = new(loc, drop_amount)
+		to_chat(user, span_notice("You have withdrawn [drop_amount] dollars."))
+		try_put_in_hand(cash, user)
+		amount -= drop_amount
+		total_stored_cash -= drop_amount
```
chazzyjazzy: this is fine, just be aware that capo and prince and a few other head honcho roles regularly spawn with >$5000, capo in particular, since i've been playing it alot, there are very few rounds where I spawn with less than $10,000

so ur options are lower all the money across the board, which id like
raise ATM total stored cash on init, which im not sure i like
or keep it how it is and have people run around to several ATMs, which i like

ill leave it up to u! also maybe finance stat on the cards starting wealth would be cool. ill leave it up to u what the calculation is

modular_darkpack/modules/economy/code/atm.dm
```diff
@@ -1,152 +1,148 @@
...
+
+/obj/item/circuitboard/machine/atm
+	name = "\improper ATM Machine"
+	greyscale_colors = CIRCUIT_COLOR_GENERIC
+	build_path = /obj/machinery/atm
+	req_components = list(
+		/obj/item/stack/sheet/glass = 1,
+		/datum/stock_part/servo = 2,
+	)
```
chazzyjazzy: there could be an exploit here. build atm, rob/deconstruct, collect money, repeat?

modular_darkpack/modules/economy/code/atm.dm
FalloutFalcon: If we raise it any higher we also make it way more profitable to smash atms.

The intent is to make it a little harder to make all your cash liquid really fast. There is kinda a similar thing in backgrounds (for htr atleast) where the `resources` background has most of your assets in non-liquid form and you have to liquidate them before its super usable.

XeonMations (approved):

Issue #439, opened 2025-12-13
TGUI error on stats menu


chazzyjazzy:
### Issue Summary:

When opening the stats menu there is a TGUI error

I'm under the assumption that this happens because of stat mods associated with Celerity, Potence and Fortitude which push the attributes past their max score, causing a display error.

Proved my hypothesis by opening the stats menu on a human. All the stats are fine. Opened it on a Toreador character and was presented with the error. (Celerity)

<img width="920" height="770" alt="Image" src="https://github.com/user-attachments/assets/92d9e354-e98c-45bf-8dfd-bfec6a840e78" />

### Reproduction Steps:

1. open stats menu with a human
2. opens fine
3. make it a toreador
4. open the stats menu
5. observe runtime

PR #440, opened 2025-12-14
Fixes TGUI stats error

XeonMations:
## About The Pull Request
Fixes https://github.com/DarkPack13/SecondCity/issues/439

image

image

## Why It's Good For The Game
no bluescreen good

## Changelog
:cl:
fix: Fixed a bluescreen caused by having too large a score in a stat.
/:cl:


chazzyjazzy (approved):
I was gonna say something akin to "Wont this allow people when limited dots are introduced to level to 10/10 even if they dont have max dots in the respective discipline" but that wont be an issue I don't think since its only taking the bonus score, so someone w/ 1 in celerity will see 6 dots on the sheet not ten

PR #451, opened 2025-12-20
Imports and reworks morality paths

chazzyjazzy:
## About The Pull Request

this pr will
[ check ] convert adjustmorality proc on mob/living to a signal handler using Niobe's signal framework
[ check ] import TFN's varied morality paths including Code of Samiel, Path of Bones, Path of Metamorphosis, and more
[ check ] change all humanity losses to a conscience (humanity alignment) roll at difficulty 6 and all path losses to a conviction roll at difficulty 6 (hierarchy of sin and other mechanics should ideally be worked on in a future pr. this one is mostly foundational steps + moving it to st stats)
[ ty xeon ] reimplement frenzying as a self control  roll and probably make it function again hopefully and retick the files since i think we have still yet to implement this
[ check ] changes a character's starting morality path score to be a sum of the virtues their path is made up of (for example, humanity is conscience + self control, so if i had 3 dots in each, id start the round with 6 humanity, same with the next round)
[ check ] change humanity to use the storyteller stat system 
[ check ] implements a 'hunger' feature where if Kindred are below 7 minus self control bloodpoints they make a roll when feeding to either enter frenzy and drain their target until they reach full bps or until the target dies
[ check ] and probably reintroduce torpor timers in some muted way (probably increasing/decreasing by 1 minute around 7 morality path score as a baseline) so that players have an incentive to care about this system

in the future the goal is to weave this into faction alignment on the preferences screen so that we can gatekeep enlightenment more so that it isn't a blatant 'morality path off' feature.

## Why It's Good For The Game

more immersion, more stats implementation, more gameplay elements

players should have a very real material incentive to invest into all stats, rather than  just brawl melee firearms strength dexterity stamina courage and willpower

and persistent morality path scores relies on xp as the only way to gain Humanity / Path is by spending experience on it. 

## Changelog

:cl:
code: reintroduces frenzy and morality paths, degeneration rolls, blood 'hunger', reasons to stat into virtues
refactor: moved humanity and morality to storyteller stats system

/:cl:


chazzyjazzy:
> seems fine, dont love how we have a separate choiced pref that still interfaces with storyteller stats tho.

Wdym?

edit : why did the bot just request a review...

PR #456, opened 2025-12-22
Inhalers in Pharmacy Shops

Icarus-The-Sun:
## About The Pull Request
Adds inhalers and refill containers to pharmacy shops, allowing anyone with the asthma quirk to not constantly end up dead.
## Why It's Good For The Game
QOL if we decide to keep the Asthma negative Quirk. Offers a way for asthmatic characters to not just die.
## Changelog
:cl:
add: Added Asthma inhalers and refill canisters for them to pharmacy shops.
/:cl:


FalloutFalcon (approved):
I think a fair few of them (like asthma) are fine and could totally but homebrew merits, I think adding an inhaler item to the store is fine regardless of if we keep the quirk.

PR #467, opened 2025-12-30
Reimplements the Penumbra

FalloutFalcon:
## About The Pull Request
re-implements the penumbra while cleaning up some of the code.
Its not considering an outside area so thats its lit with starlight, meaning it will replicate some of the state of the real world.

Pretty bare bones, I will want to revisit it to give it unique game-play like banes and umbra fishing

requires #504

## Why It's Good For The Game
werewolf content.

## Changelog
:cl:
fix: fixes some issues with umbra background not appearing in the umbra
qol: some qol for transfer points including allowing ghosts to use them
/:cl:

### DIFF
--- a/modular_darkpack/modules/z_travel/code/transfer_point.dm
+++ b/modular_darkpack/modules/z_travel/code/transfer_point.dm
@@ -20,7 +20,6 @@ GLOBAL_LIST_EMPTY(unallocted_transfer_points)
 
 /obj/transfer_point_vamp/Initialize(mapload)
 	. = ..()
-	apply_wibbly_filters(src)
 	if(id && !exit)
 		if(isnum(id))
 			// Im considering them bad practice because you cant tell where they lead - Fallcon


modular_darkpack/modules/umbra/code/parallax.dm
```diff
@@ -0,0 +1,36 @@
...
+	var/client/boss = source.canon_client
+	on_z_change(boss.mob)
+
+/atom/movable/screen/parallax_layer/umbra/proc/on_z_change(mob/source)
+	SIGNAL_HANDLER
+	var/client/boss = source.client
+	var/turf/posobj = get_turf(boss?.eye)
+	if(!posobj)
+		return
+	SetInvisibility(is_mining_level(posobj.z) ? INVISIBILITY_NONE : INVISIBILITY_ABSTRACT, id=type)
```
chazzyjazzy: Set invisible? Is there no better way to handle getting rid of this parallax layer?

chazzyjazzy (approved):
Soft accept because I usually dont do map stuff since thats a bit more creative, code looks fine, don't see any problems

modular_darkpack/modules/umbra/code/parallax.dm
FalloutFalcon: This is how it was handled and I think is still handled for parallax not specific to the station.

PR #497, opened 2026-01-13
Fixing Spellbook Icon States

chazzyjazzy:
## About The Pull Request

Splat rework just needed conditional logic in add_power to handle for /datum/action/discipline/path rather than /datum/action/discipline which defaults to the original set of icons and icon states rather than the path specific

## Why It's Good For The Game

aysedfgjioeoie

## Changelog

:cl:

fix: fixed spellbook logic to match splat rework 

/:cl:


FalloutFalcon (changes requested):
why not just move action_type onto the base type? less snowflakey

PR #504, opened 2026-01-13
Makes directionals sane on some city signs

FalloutFalcon:
## About The Pull Request
Moves some hidden icon states out of directionals and adds them as alt typepaths or icon states instead.
## Why It's Good For The Game
Sign code cares a lot about what direction an object has for the purpose of mapping, this means you cant stash extra states in your directions and they need to be seperate states


chazzyjazzy:
Linters

chazzyjazzy (approved):

PR #541, opened 2026-01-20
Firearm stat recoil reduction

chazzyjazzy:
## About The Pull Request

This PR makes it so that, upon firing a weapon, the user's STAT_FIREARMS skill is rolled, with each success reducing the initial recoil by 1

There is a cooldown that lasts for 30 minutes to prevent rolling for each and every time the gun is fired

## Why It's Good For The Game

more stats

## Changelog

:cl:
add: firearms stat roll to reduce recoil

/:cl:

### DIFF
--- a/code/modules/projectiles/guns/ballistic.dm
+++ b/code/modules/projectiles/guns/ballistic.dm
@@ -597,7 +597,7 @@
 	if (sawn_off)
 		bonus_spread += SAWN_OFF_ACC_PENALTY
 
-	//DARKPACK EDIT ADD - recoil
+	// DARKPACK EDIT ADD - recoil
 	if(COOLDOWN_FINISHED(src, recoil_skill_check))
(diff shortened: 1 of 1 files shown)


XeonMations:
Surely there's a way of doing this without having to add so many cooldowns and checks? maybe in whatever uses the ``recoil`` variable?

chazzyjazzy:
> Surely there's a way of doing this without having to add so many cooldowns and checks? maybe in whatever uses the `recoil` variable?

really the issue with this pr is that theres like 3 completely separate subtypes so i had to do the same thing across all 3

chazzyjazzy:
> I dont see a reason we cant just move it up to /ballistic frankly.

was also thinking that but then i got scared

will do it

chazzyjazzy:
ty xeon

FalloutFalcon (approved):

PR #545, opened 2026-01-21
Kills most door subtypes

FalloutFalcon:
## About The Pull Request
Updates path to replace all existing doors with there relevant helpers.
## Why It's Good For The Game
Helpers are much clearer and easier to audit.
## Changelog
:cl:
map: All mapped doors work off of helpers rather then varedits 
/:cl:


chazzyjazzy (approved):
/obj/effect/mapping_helpers/door/lock

is this the parent type of the lockpick difficulty maphelper or are we missing subtypes here? notice that its on basically every updatepaths

PR #550, opened 2026-01-22
Readds Generation onto preferences

chazzyjazzy:
## About The Pull Request

This PR readds generation onto preferences by commenting out the required TRAIT_DRINKS_BLOOD and adds an is_accessible() method which first checks if the prent returned false and if not checks if the species preference has /datum/species/human/kindred selected (which obviously needs to change to some kind of splat check)

## Why It's Good For The Game

bugfix

## Changelog

:cl:

fix: generation on prefs menu added back

/:cl:


chazzyjazzy:
> Cant you just add TRAIT_DRINKS_BLOOD to the species AND the splat (assuming its just because lucia moved it down to splat). There are already multiple instances of that like clan.

The problem is that splat doesnt exist til joining the round, it exists nowhere in preferences, nor on the mob/dead/observer/whatever the new player is on the title screen

I can give it to the species however but... is that wanted? seems like it was removed from species/kindred...

FalloutFalcon (approved):

Issue #477, opened 2026-01-04
You can adjust or reset your stats midround via prefs


FalloutFalcon:
### Issue Summary:

You can reset or change your stats midround at will. 
The list between your pref and your mob are not mean to be linked, but are made such because it directly `Copy`s the list 

<img width="664" height="87" alt="Image" src="https://github.com/user-attachments/assets/0f353f2e-794d-442f-b31f-16e54c075266" />

### Reproduction Steps:

I read a book. went "ragh i need to bump my stats" went into my prefs, increased occult, cracked open another book and i was suddenly rolling with 5 dice now.

Faengi:
This was resolved in https://github.com/DarkPack13/SecondCity/pull/551 wasn't it?

PR #69, opened 2025-08-23
Stat preferences

XeonMations:
## About The Pull Request

adds v20 stats to the game

## Why It's Good For The Game

stats good

## Changelog

:cl:
code: Added Stats to the game
/:cl:

PR #551, opened 2026-01-22
Fixes minor character sheet exploits or oversights.

FalloutFalcon:
## About The Pull Request
Prevents people from adjust their stats, or swapping there slot to a different character, without first returning to the lobby.
image
## Why It's Good For The Game
The character slot one could prob also verify that said character is spawned in round, this is a light bandaid.
Same story with stats but also this only exists due to your stats list being shared by both your character and pref sheet which is an oversight caused by #69. This "fixes" #477 without addressing the core issue causes it.
## Changelog
:cl:
fix: fixes swapping stats around while spawned in
fix: prevents swapping character slots to avoid stuff like masquarde being dinged to the wrong character
qol: bonus die on character sheet is displayed in white
/:cl:


chazzyjazzy (approved):
awesome

PR #631, opened 2026-02-06
Improvments to destruction behavoir and other things of gates

FalloutFalcon:
## About The Pull Request
Doing work on totems, I realized we needed a more generic way to say "we have atom integ but please dont acctually destroy us". See fuseboxes, totems, and the gate. Where we dont have ways to replace them ingame nor should we allow it to be broken.

Also took the time to add a quaint little graveyard in runtime town. The zombies already only bother you if you spawn as a gravekeeper so its useful to have. Made a rapidly spawning grave so you dont have to wait as well.

image

## Why It's Good For The Game
Much less snow-flaking required to achieve the desired behavior

## Changelog

:cl:
fix: tiles next to the graveyard gate are made dense automaticly
refactor: Gravekeeper gate should behavior more consistently in line with atom integrity
/:cl:

### DIFF
--- a/code/game/atom/atom_defense.dm
+++ b/code/game/atom/atom_defense.dm
@@ -10,8 +10,10 @@
 	var/integrity_failure = 0 //0 if we have no special broken behavior, otherwise is a percentage of at what point the atom breaks. 0.5 being 50%
 	///Damage under this value will be completely ignored
 	var/damage_deflection = 0
+	// DARKPACK EDIT ADD START
 	/// Atom uses integrity but will not be deleted upon reaching 0 through normal means
-	var/prevent_destruction = FALSE // DARKPACK EDIT ADD
+	var/prevent_destruction = FALSE 
+	// DARKPACK EDIT ADD END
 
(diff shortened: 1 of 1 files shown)


code/game/atom/atom_defense.dm
```diff
@@ -10,6 +10,8 @@
 	var/integrity_failure = 0 //0 if we have no special broken behavior, otherwise is a percentage of at what point the atom breaks. 0.5 being 50%
 	///Damage under this value will be completely ignored
 	var/damage_deflection = 0
+	/// Atom uses integrity but will not be deleted upon reaching 0 through normal means
+	var/prevent_destruction = FALSE // DARKPACK EDIT ADD
```
XeonMations:
```suggestion
	// DARKPACK EDIT ADD START
	/// Atom uses integrity but will not be deleted upon reaching 0 through normal means
	var/prevent_destruction = FALSE 
	// DARKPACK EDIT ADD END
```

XeonMations (changes requested):

Marshmellow105 (approved):
very cute graveyard

PR #640, opened 2026-02-07
Whispering in Obfuscate wont break it

FalloutFalcon:
## About The Pull Request
The book describes ofbuscate as being broken per person unless its an obv loud action like yelling. Thus it follows that whispering in obf would only break it to the person that hears it. Which is missing in the current implementation but this allows more more roleplay.

## Why It's Good For The Game
Half ttrpg accurate pending a more full rework
## Changelog
:cl:
balance: whispering does not break obfuscate
/:cl:

### DIFF
--- a/html/changelogs/AutoChangeLog-pr-681.yml
+++ b/html/changelogs/AutoChangeLog-pr-681.yml
@@ -1,0 +1,4 @@
+author: "chazzyjazzy"
+delete-after: True
+changes:
+  - rscadd: "Readds character height customization as a numerical preference slider"
--- a/modular_darkpack/modules/preferences/height_preference.dm
+++ b/modular_darkpack/modules/preferences/height_preference.dm
@@ -1,0 +1,25 @@
+/datum/preference/numeric/height
+	category = PREFERENCE_CATEGORY_NON_CONTEXTUAL
+	savefile_key = "height"
+	savefile_identifier = PREFERENCE_CHARACTER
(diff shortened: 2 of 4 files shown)


XeonMations:
Obfuscate 2 states the following: "Even a whispered word or the scuffing of a shoe against the floor can be enough to disrupt the power."

"Other acts may require a certain number of successes; speaking quietly without giving away one’s position, for instance, demands at least three successes"

The above also applies for obsucate 4 and 5, due to "When fully invisible, the vampire is handled as described under Unseen Presence, above."

FalloutFalcon:
> Obfuscate 2 states the following: "Even a whispered word or the scuffing of a shoe against the floor can be enough to disrupt the power."
> 
> "Other acts may require a certain number of successes; speaking quietly without giving away one’s position, for instance, demands at least three successes"
> 
> The above also applies for obsucate 4 and 5, due to "When fully invisible, the vampire is handled as described under Unseen Presence, above."

The text to me and others Ive read it with seem to imply that breaks of obfuscate do not necessarily cancel the power outright but rather break the mental effect for whomever witnesses it. Otherwise the means existance of person with auspex in the room (including your own party members) would mean your constantly rolling to keep the power up to ANYONE and would make obfuscate far far worse with that interpretation. Because of this I want to do a more full rework, but the inability to communicate with players even a little makes this power feel very anti role-play as you quite literately can only sit their and observe.

Searching the web and alot of people ask similar questions and most everyone else seems to say yes. 
image

chazzyjazzy:
so whats the consensus

PR #655, opened 2026-02-08
Makes typhonic beasts break the masquerade

XeonMations:
## Why It's Good For The Game
Giant snake. Need I say more?

## Changelog
:cl:
fix: Typhonic beasts now break the masquerade.
/:cl:


FalloutFalcon (approved):
Still feel like we need a better/more standard method of doin this.

PR #713, opened 2026-02-24
The Automaticening

SELFHELL:
## About The Pull Request

Adds the Automatic Fire component to most Full-Auto firearms.

Given our extremely asymmetrical balancing, this is going to need some extensive in-round testing and adjustments to catch big oversights I'd not found in my own testing.

## Why It's Good For The Game

My Unregistered Automatic Weapon Should Not Be Shooting In A 4 Round Burst Each Click 
## Changelog
:cl:
add: Most Automatic Guns have gotten Full-Auto Capacity, please make an issue report for extremely unfair TTKs
/:cl:

### DIFF
--- a/modular_darkpack/modules/weapons/code/guns.dm
+++ b/modular_darkpack/modules/weapons/code/guns.dm
@@ -313,7 +313,7 @@
 	rack_sound = 'sound/items/weapons/gun/pistol/slide_lock.ogg'
 	fire_sound = 'modular_darkpack/modules/weapons/sounds/mac10.ogg' // DARKPACK sound effect
 	serial_type = "GBI"
-	var/rof = 0.05 SECONDS //1200 RPM, if any other gun has the same fire rate as this gun, increase this gun so its the new fastest
+	var/rof = 0.055 SECONDS //1090 RPM, if any other gun has the same fire rate as this gun, increase this gun so its the new fastest
 
 /obj/item/gun/ballistic/automatic/darkpack/mac10/Initialize(mapload)
 	. = ..()


modular_darkpack/modules/weapons/code/guns.dm
```diff
@@ -307,6 +313,11 @@
 	rack_sound = 'sound/items/weapons/gun/pistol/slide_lock.ogg'
 	fire_sound = 'modular_darkpack/modules/weapons/sounds/mac10.ogg' // DARKPACK sound effect
 	serial_type = "GBI"
+	var/rof = 0.05 SECONDS //1200 RPM, if any other gun has the same fire rate as this gun, increase this gun so its the new fastest
```
FalloutFalcon:
```suggestion
	var/rof = 0.055 SECONDS //1090 RPM, if any other gun has the same fire rate as this gun, increase this gun so its the new fastest
```
.45 is seemingly this rpm? citation needed. I would also say we should lean towards a lower rpm because god is this probably too much damage to be fair loo.

this seems.... a bit unfun.

https://github.com/user-attachments/assets/f3a8e65c-12e9-4f0f-892c-3020b1c00e7c

amour seems to make it... survivable somewhat... I think this is mostly just cause i didn't roll enough accuracy on my roll. It also seems a bit laggy? The lag seems to be enough to cancel my click.

https://github.com/user-attachments/assets/dafadf32-074b-4109-b547-0e44a15c1857

But I look at this and just... man.

https://github.com/user-attachments/assets/9903e547-5378-4f99-a10a-ab72d294e5f7

SELFHELL: Yeahhhhh... that was one of the ones I didnt play with too much.. mostly focused on getting the MP5/AK74 Good..

FalloutFalcon:
> code looks fine as long as you address Abby's concern

I realize the mac10 is not the default anarch gun so its.. mostly fine for now

PR #737, opened 2026-02-27
Fetch enlightenment directly from path

FalloutFalcon:
## About The Pull Request
Creates a helper proc to fetch enlightenment from the stat and removes the enlightenment var on splat.
## Why It's Good For The Game
More consistent to just directly check a datum rather then the datum having to set a var on a mob. which could be desyned if something else (like the new for kindred splat) sets it.
## Changelog
:cl:
code: Fetches enlightenment directly from Path.
/:cl:

### DIFF
--- a/modular_darkpack/modules/vampire_the_masquerade/code/preferences/morality/morality.dm
+++ b/modular_darkpack/modules/vampire_the_masquerade/code/preferences/morality/morality.dm
@@ -41,7 +41,7 @@
 	stat_morality.morality_path = new value(target)
 
 /mob/living/proc/is_enlightenment()
-	var/datum/st_stat/morality_path/morality/stat_morality = storyteller_stats["[STAT_MORALITY]"]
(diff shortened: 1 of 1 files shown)


modular_darkpack/modules/storyteller_stats/code/st_stats/base_type/_morality.dm
```diff
@@ -1,8 +1,8 @@
 /datum/st_stat/morality_path
 	abstract_type = /datum/st_stat/morality_path
 	category = "Morality Path"
-	starting_score = 2
+	starting_score = 7
```
chazzyjazzy: this breaks stuff because a players starting morality score i've been told is the sum of the morality path's relevant virtues in the TT, and every player starts with 2 ( 1 each) in conscience and self control so morality path starts with 2

chazzyjazzy (changes requested):

PR #738, opened 2026-02-27
implements permanent ghouling and embracing

lpeapnni:
## About The Pull Request
Reimplements the code to handle permanent embracing/ghouling. As of right now, the discipline bits still need to be implemented.

embracing:
image
image

ghouling:
image
image
## Why It's Good For The Game
Bro
## Changelog
:cl:
code: Reimplemented permanent ghouling and embracing
/:cl:


lpeapnni:
> looks fine im just wondering if its the case that the childe's morality path and score in it is the same as the sire's. kinda breaks things too considering a player's morality path score is the sum of their relevant virtues (for ex, humanity has conscience and self control, so 3 in each gives humanity 6), which may deviate from the sire's selected virtues.

just gave it a quick test; the morality path is the same as the sires (as intended), but the score isnt the same. is the score supposed to be the same as the sires?

modular_darkpack/modules/vitae/code/ghouling.dm
```diff
@@ -29,7 +30,7 @@
 	if(master != owner)
 		mind.enslave_mind_to_creator(owner)
 		apply_status_effect(/datum/status_effect/blood_bond, owner)
-		to_chat(src, span_userdanger("<b>From the first sip from the first night, you shall idolize or hate from whom you've drank deep. From the second sip upon the second night, you shall  likewise hold onto their words and seek to curry their favor. Upon the third night and third drink, you are utterly theirs. Acts of defiant will are possible, but they are fleeting.</b>"))
+		to_chat(src, span_userdanger("<b>From the first sip from the first night, you shall idolize or hate from whom you've drank deep. From the second sip upon the second night, you shall likewise hold onto their words and seek to curry their favor. Upon the third night and third drink, you are utterly theirs. Acts of defiant will are possible, but they are fleeting.</b>"))
```
buffyuwu: nitpick, but the userdanger class should already be bold as per:

`
.userdanger {
  color: hsl(0, 73.6%, 44.5%);
  font-weight: bold;
  font-size: 185%;
}
`

PR #743, opened 2026-03-06
Makes indexes of stat lists a type rather then string

FalloutFalcon:
## About The Pull Request
Makes indexes of stat holder lists a type rather then string
## Why It's Good For The Game
The only place a string was useful was tgui so we do text2path there instead of EVERYWHERE else.

Allows for better handling, as you can access the type directly if you need something like a name, and dont have to wrap all logic on strings which is annoying.

Its a hanger-on from when we were direct saving the list for prefs. this is no longer needed.
## Changelog
:cl:
code: List index for stats was made its type instead of a type as a string
/:cl:


FalloutFalcon:
Im realizing this seems to be a breaking change for pref. and i will prob want to write an update script for it....
plus imma wait on #738

PR #747, opened 2026-03-11
Fixes eyes for kiasyd and salubri and removes todo comments for gridventory

lpeapnni:
## About The Pull Request
Gives Kiasyd and Salubri their eyes back. Also fixes a bug where the game would throw a warning over on_join_round getting overwritten. ALSO also removes some todo comments relating to gridventory stuff because I'm pretty sure we gave up on that
image
image
## Why It's Good For The Game
👀
## Changelog
:cl:
fix: fixes eyes for both kiasyd and salubri
code: removes todo comments for gridventory stuff
/:cl:


modular_darkpack/modules/splats/code/subsplat/_subsplat.dm
```diff
@@ -27,7 +27,7 @@
 
 	// Applies on_join_round effects when a client logs into this mob
 	if(joining_round)
-		RegisterSignal(gaining_mob, COMSIG_MOB_LOGIN, PROC_REF(on_join_round))
+		RegisterSignal(gaining_mob, COMSIG_MOB_LOGIN, PROC_REF(on_join_round), override = TRUE)
```
FalloutFalcon: If this is being hit it means on_gain is being called on a mob twice (or being registered to two different mobs. I forget rn) or the signal is not being unregistered. This is its own issue that should not be patched over by using override (which is generally bad and causes some weird behavoir)

PR #750, opened 2026-03-12
Giovanni Job Titles & Clan Availability Reduction, Tremere/Sabbat Titles

Biplume:
## About The Pull Request

La Squadra replaced with 'Soldato'
La Famiglia replaced with 'Socio' and closed to Kindred other than Caitiffs
Chantry Archivist replaced with Chantry Apprentice with Chantry Acolyte and Chantry Magister alt-titles.
Sabbat Pack replaced with True Sabbat
## Why It's Good For The Game

I tried to find other examples of clans being within the Giovanni pre-Hecata in lore and couldn't find anything. I don't think it thematically makes sense to have Brujah/Ventrue/Nosferatu/Gangrel/Toreador/Malkavian/Lasombra Giovanni given their whole bit of leaning into being an independent family that is so closed off that they can be described as 'incestuous' both figuratively and literally. The only example I could find at all was Alfie Rossellini who was shovelheaded by a Lasombra in V20 Ready-Made Characters and this character seems to have been othered by the family and treated as an in-between rather than a family member for it. It's also the only example I found at all. Caitiffs sometimes slide in I was told, though I haven't had an exact citation for this yet.

The Giovanni roles were renamed because the old names were literally a JJBA reference. The replacements are not with citation  because I couldn't find ranks for regular family members and so I crowdsourced these from mafia terms since half the higher ones are anyway.

Sabbat Pack was replaced with True Sabbat from V20 Companion referring to ordinary full Sabbat members.

Archivist was replaced with actual Tremere ranks from Lore of the Clans since the other names all use their rank rather than their cover job.
## Changelog
:cl:
:typo: La Squadra renamed to Soldato/a
:typo: La Famiglia renamed to Socio/a
:typo: Chantry Archivist renamed to Chantry Apprentice
:typo: Sabbat Pack renamed to True Sabbat
:add: Chantry Acolyte and Chantry Magister alt titles added to Chantry Apprentice
:del: All clans except Caitiff removed from Socio/a
/:cl:

### DIFF
--- a/tools/UpdatePaths/Scripts/DarkPack/748_job_landmarks.txt
+++ b/tools/UpdatePaths/Scripts/DarkPack/748_job_landmarks.txt
@@ -1,4 +1,0 @@
-/obj/effect/landmark/start/darkpack/chantry/archivist : /obj/effect/landmark/start/darkpack/chantry/apprentice
-/obj/effect/landmark/start/darkpack/hecata/famiglia : /obj/effect/landmark/start/darkpack/hecata/soldato
-/obj/effect/landmark/start/darkpack/hecata/squadra : /obj/effect/landmark/start/darkpack/hecata/socio
-/obj/effect/landmark/start/darkpack/sabbat/pack : /obj/effect/landmark/start/darkpack/sabbat/true_sabbat


LucifersCoils:
I'll PR this another time less half-baked. Not my best work.

PR #780, opened 2026-03-19
[IDB IGNORE] TFN Playtest Fixes/Updates

buffyuwu:
## About The Pull Request

contains fixes and updates from the final nights' most recent playtest

## Changelog
:cl:
add: adds much more variety to examine flavortexts for all (and lackthereof) gender(s)
add: Adds more hallucinations to malkavians! look out for your mother to scold you, with different scolding phrases depending on your immortal age. e.g. milennial vamps get told to stop watching TV, 1920s vamps get told to go tend to the horses
refactor: greatly increased the time between npcs stopping to talk after being bumped/grabbed and them starting to move again. this should make it easier to pull one away to feed on. TODO: refactor it completely so that it's more interactive for kindred
refactor: no longer uses a lot of the vtm13 era glassware that have static names and icons. with the modern tg glassware, drinks will automatically change glasses to the appropriate ones for the drinks you mix into them
refactor: derangement hallucination speech is easier to distinguish and uses more artistic styling
refactor: refactors discipline middleware a bit. you will now be prompted if you try to join with 0 dots assigned, and if you try to change age you will have to wipe your discs.
refactor: phone sound volume for keyboard clicks have been increased, and apps now play a sound when you open them, too. phone navigation should thereby be audible to other players, making it harder to hide that you're on your phone
bugfix: trusted status now saves from the secondcity disc admin menu when set/unset
bugfix: your generation now properly applies. previously everyone was silently defaulted to gen 10 due to the pref not being fully implemented
bugfix: endbrowser navigation buttons now work on every endbrowser screen as intended
bugfix: phones no longer display new text notifications twice, and the simulated delay between receiving the text and the notification now correctly fires. this delay is intended to give the phones a 'realistic' carrier lag time between when the text arrives and when you get notified for a new text, just like real phones
bugfix: fixed missing sprite for broken tube lights
bugfix: fixed missing sprite for sheaths
add: added a warning if you try to join a round before assigning disc dots. this should help new and/or excited players from accidentally spawning in with 0 disciplines. you can choose to accept the warning and continue with just the clan commons; rares must always be assigned through the menu or via an admin
bugfix: @XeonMations fixed kiss status effect for mortals being infinite. now it will fade after 30 seconds. avoids requiring excessive logic (previously the idea was to have it be a status effect WHILE being bitten, but this is unnecessary and not performant. a simple expire time will do)
bugfix: fixes satchels showing an error sprite when concealed
bugfix: you can now use boozeomats and they no longer shout space station slogans at you
[...]

PR #804, opened 2026-03-25
Minor revert of two changes from #780

FalloutFalcon:
## About The Pull Request
reverts the drink and boozeomat changes made in #780 as they were never addressed and pushed through.
## Why It's Good For The Game
The removal of all drink glasses expect wine_glass (which also put it outside of the modular edit) was not justified or really explained.

For the init code, we have a type that is better to use and its bad to have random setting initialize behavior like that as it overides subtype and varedit behavoir.
image


buffyuwu (dismissed):
above

FalloutFalcon:
> it'll still yap station slogans unless you null those out, we need the change. if people want snowflakey machines that cost credits, theyll need to make a credits system.

The vending machines already use dollars and credit cards as i built out the code to automaticly support them in most places. 

> if they want to have one that yaps slogans, they can override the init proc

The init override was a poor way to adress the yapping issue. the slogans are also displayed in the TGUI which is flavorful and we should not remove.

> the commented out wod13 glasses can be outright removed, they dont update sprites nor name when drinks are poured in and are mostly broken as-is; i think the shot glass straight up vanishes when you pour stuff into it

Then they should prob be fully removed, however i really like how they look. They fit a lot better in the art style and I am leaning towards reusing the sprites of Collins and shotglass for the tg types for the but the other ones can just exist tbh. Asking for something to be served in a pint is flavor, and they have unique reagent sizes which i think is neat.

>also i think there are higher priority items to worry about than this :(

I see bad code, i fix bad code.

chazzyjazzy:
does this still need to be open? @FalloutFalcon idk whats the status here

FalloutFalcon:
> does this still need to be open? @FalloutFalcon idk whats the status here

I already stated how my changes already fixed buffys requests, so i dont super see a reason not to remove the redundant init overrides.

chazzyjazzy (approved):
i have no clue what was contentious about this PR, its been sitting here for so long that we have a long-fixed bug being talked about in the comments

PR #810, opened 2026-03-28
Gargoyle wings & flight

chazzyjazzy:
## About The Pull Request

This PR gives Gargoyles their wings sprite overlay and allows them to use a /datum/action to fly. I originally thought this was way harder than what it was because I didn't know about the wings/functional subtype and sprite_accessory datums. when I found those this PR ended up becoming like, forty lines lol.

## Why It's Good For The Game

Gargoyle

## Changelog

:cl:
add: Adds Gargoyle wings and flight
/:cl:


buffyuwu (dismissed):
please make these very slow and require a 5 second do_after to move upwards or downwards while flying. make them exit flight mode when they take damage as well. we want to get ahead of the z-level spamming flying rat archetype

chazzyjazzy:
> this is probably where I put on my nerd glasses and say that gargoyles get only one dot of flight free which only allows you to soar, not fly, and that being able to do a takeoff is later on if we want to have that be a limiting factor (e.g. dot 1 could just cut fall damage rather than give you actual z-flight, giving it an investment cost as an action)
> 
> image image

i have no interest in implementing this.

FalloutFalcon (approved):

PR #847, opened 2026-04-02
Obtenebration 5 Fix

Kolibri-B:
## About The Pull Request

The rebase broke Obtenebration 5's ability to phase through doors. This fixes that and adds a few more object types that an amorphous shadow-blob should be able to phase through.

## Why It's Good For The Game

It's a bugfix, and also makes the ability more lore-accurate.

## Changelog
:cl:
add: Added a few more objects that Obtenebration 5 can phase through
fix: Fixes Obtenebration 5 not phasing through doors
/:cl:


dwinters99 (changes requested):
comment sanity: this is darkpack, proper module notation

dwinters99 (dismissed):

dwinters99:
> ...do things such as pass through trees, rocks, and possibly some other large objects that I thought may be questionable... 

This is pretty questionable... but I don't know if PASSMACHINE is the answer. I'll leave that decision to @FalloutFalcon.

FalloutFalcon (changes requested):
remove instances of `TRAIT_PASSDOOR` since it was only used for this.

I really should have passstructure as that has a buncha stuff you would expect to be able to move under, and therefore its fine to keep a bunch of these as PASSSTRUCTURE rather then PASSMACHINE

FalloutFalcon (changes requested):
you still need to remove all instances of TRAIT_PASSDOOR!

Image

Kolibri-B:
> you still need to remove all instances of TRAIT_PASSDOOR!
> 

oh, I misunderstood this, sorry. nevermind, one sec.

FalloutFalcon (approved):

chazzyjazzy (approved):
yes, kolibri, YES!

PR #871, opened 2026-04-05
Kills "ALL" defines for clans and splats

FalloutFalcon:
## About The Pull Request
Replaces them with just null as a var because why would you check something against a list if its meant to always be there.
## Why It's Good For The Game
These only exist to allow oversight and bad behavoir (i.e someone adding a funny bloodline but since it was not added to CLAN_ALL its not usable anywhere we use it)
we have perfectly good global of all clans and splats if you need to interrate through every ID.
## Changelog
:cl:
code: Restructures code to prevent oversights from failure to add clans and splats to a list
/:cl:


chazzyjazzy (approved):
looks good

PR #873, opened 2026-04-05
Obfuscate & Presence Changes

Stutternov:
## About The Pull Request

This is a PR that focuses on the following changes; most being minor but semi-flavorful. I've broken them down by section below.

### Presence Changes

Attempted to mimic Awe, Dread Gaze, and Entrance from V20.

**Awe** has received the ability to burn one WP to resist out of it, akin to tabletop where you can spent 1 WP to burn away.
image

**Dread Gaze** has received a semi-face lift; re-adding the 'move away from' effect on 3 success. Dread gaze has also gained the ability that, if you have greater than or equal to as many successes as the target has dice, you get the extended action of making them 'cower' (Attempts to do V20's thing of forcing them into a ball on an extended action; instead stuns, trembles, and debuffs their dex-movement hard.)

**Entrance** has received the same thing as Awe, however you must roll your willpower against a difficulty of 8 to break from it. If you don't succeed, won't be able to resist again. Burns a WP point still.

__Note:__ Dread Gaze is 'working' but has issues with moving currently. It.. tends to teleport people instead of force them to flee, annoyingly.

### Obfuscate Changes

Adds in a new check for if you slip or step on a loud object just like V20 rules have it. If you do, it's your wits and stealth against a difficulty roll to stay stealthed.

Just as tabletop has it, if you step on a loud object its a difficulty 5 check - relatively easy-ish to pass. Slip? Difficulty 9 check.

This gives some flavorful counter play to obfuscated users and makes more use of the skills.

__Note:__ I have a video of this one working but it won't let me upload to Github.. annoying.

Also, it is messy how the signal was implemented. Feel free to suggest a better place to shove that signal, I could not make it modularized without tons of errors.

### Dominate Changes

I was going to more rework conditioning but I decided to hold off since it would be a much larger change. Instead just fixed Dom 3's wording for now to fit the table-top successes.

## Why It's Good For The Game

**Presence** really benefits from having dread gaze be slightly more useful both mechanically and roleplay wise, given it's relatively useful on TT and should reflect that. Plus, I personally like that it makes people not treat wits as a dump-stat. Also Awe and Entrance having the ability to resist out of in some way is a bit nice given how long they can last or such; plus it's accurate to TT rules. 

**Obfuscate** is cool but I felt the checks are good to have on an otherwise passive power that has no upkeep cost etc. This way you have a few ways of breaking out of obfuscate if you don't watch where you're stepping, move obsticals, or choose other routes. 

**Dominate** is relatively fine, though should be kept accurate to TT. Future-wise I wanted to add ability to restore memories via the checks Dominate 3 has on TT to restore memories, as well as make conditioning be the one to change personalities to be malleable (3 doesn't inherently, but 4 does) , but I'll likely do that separate. This just makes sure dominate is accurate to how it is in V20's rulebook. (Using V4 from 2011)

**Optimally these will be better with the blood-cost change PR** - as dominate is free BP wise (just like TT) and so is Dread Gaze, etc.

## Changelog
:cl:
add: Added new Obfuscate checks for stepping on new loud objects (glass, sticks, lightbulbs, etc) and slipping)
add: Added resist button for Awe and Entracement burning WP
add: Added new Dread Gaze interaction forcing targets to flee, as well as extended action option on large success
del: Removed Success 6 off Dom and redid some wording to be more accurate to TT description.
/:cl:


code/game/objects/items/stacks/sheets/glass.dm
```diff
@@ -404,6 +404,7 @@ GLOBAL_LIST_INIT(plastitaniumglass_recipes, list(
 		var/mob/living/L = AM
 		if(!(L.movement_type & MOVETYPES_NOT_TOUCHING_GROUND) || L.buckled)
 			playsound(src, 'sound/effects/footstep/glass_step.ogg', HAS_TRAIT(L, TRAIT_LIGHT_STEP) ? 30 : 50, TRUE)
+	SEND_SIGNAL(AM, COMSIG_MOB_LOUD_STEP)	//DARKPACK EDIT - Discipline Checks (Obfuscate)
```
chazzyjazzy: indent line 407. You realize you're sending that signal even if mob/dead (observers) cross the glass shard?

modular_darkpack/modules/powers/code/discipline/obfuscate/obfuscate.dm
```diff
@@ -58,6 +58,32 @@
...
+	else
+		to_chat(owner, span_danger("Your Obfuscation falls away as you reveal yourself!"))
+		try_deactivate(direct = TRUE)
+		deltimer(cooldown_timer)
+		cooldown_timer = addtimer(CALLBACK(src, PROC_REF(cooldown_expire)), COMBAT_COOLDOWN_LENGTH, TIMER_STOPPABLE | TIMER_DELETE_ME)
+
+/datum/discipline_power/obfuscate/proc/on_slip(datum/source)
+	SIGNAL_HANDLER
+
+	var/roll = SSroll.storyteller_roll(owner.st_get_stat(STAT_WITS) + owner.st_get_stat(STAT_STEALTH), 9, owner)
```
chazzyjazzy: datumize rolls

chazzyjazzy (changes requested):
youre supposed to spend willpower to resist as you stand in the presence user's presence. 

what this means is this is not valid :

I use presence 1 on the cops and say 'let me out' and they resist and go 'no!' and then stand there continuing to interrogate me

it would be more like i use presence 1 on the cops and say 'let me out' and they resist and go 'no!' and if they continue to stand there to interrogate me, theyd be burning wp the whole time.

PR #879, opened 2026-04-06
Adds a bunch of bloodlines & changes clan icons

Biplume:
## About The Pull Request

I'm adding a bunch of bloodlines that don't require mechanics. The selection is for that reason.
Dominate Malkavians are the same but sub one discipline.
Warrior Setites are the same but sub one discipline.
Ventrue antitribu are the same but sometimes sub one discipline. They are listed here under the assumption they have done the relevant ritual - not a ton of point otherwise, right?
Tlacique are the same but sub one discipline (which counts as in-clan for them) and technically also a Thaumaturgy path that is out-of-clan. Somewhat tenuous as a result.
Banu Haqim Vizier have an RP flaw, and their Quietus variant is an optional rule from Lore of the Clans. Most tenuous as a result.
Samedi, Harbingers, Salubri Warriors, Followers of Set get icons based on their pre-V5 state. The first three are replacing codersprites, and the fourth is because we don't seem to use the Ministry.
Sprites by @Rosychu 
## Why It's Good For The Game

More bloodlines is honestly not fully necessary for some of them, but why have admins spend time doing discipline swaps when you could simply not?
## Changelog
:cl: LucifersCoils, rosychu
add: Added Banu Haqim Viziers, Dominate Malkavians, Tlacique, Ventrue antitribu, and Warrior Setites to clan selection
image: Replaced the Harbingers of Skulls, Followers of Set, Samedi, and Salubri Warriors icons
/:cl:


modular_darkpack/modules/jobs/code/camarilla/primogens/banu_haqim.dm
```diff
@@ -26,7 +26,7 @@
 	minimum_immortal_age = 50
 	minimal_masquerade = 5
 	allowed_splats = list(SPLAT_KINDRED)
-	allowed_clans = list(VAMPIRE_CLAN_BANU_HAQIM)
+	allowed_clans = list(VAMPIRE_CLAN_BANU_HAQIM, VAMPIRE_CLAN_BANU_HAQIM_VIZIER)
```
FalloutFalcon: Im noticing less addtions to job defines then i might assume was needed in this pr. Have to acctually checked to see how many of these should appear in stuff like the 3 major vampire faction jobs?

Image

FalloutFalcon (changes requested):
They also all are missing entires in the bloodwalk code but also that really ought to be a var on the splat anyway to avoid every bloodline needing to be added here like this. 

Image

modular_darkpack/modules/jobs/code/camarilla/primogens/banu_haqim.dm
chazzyjazzy: this btw. with ur changes, banu haqim vizier can only play primogen. best to just inherit the parent's ID? idk.

FalloutFalcon:
This PR overlaps with #863, since it has its own iconstates it probally supercedes it.

dwinters99 (approved):
Approval on behalf of @Major00 **for the icons.**

FalloutFalcon (approved):

PR #882, opened 2026-04-06
Adds Weapon Weights to Darkpack Guns

Stutternov:
## About The Pull Request

Title really says it all.

I was wondering why AR-15s, AKs, etc could be fired akimbo all the time yet shotguns couldn't - then realized no weapon weights were ever set for the base Darkpack firearms.

Does the following:
- SMGs and AR's all now have medium weapon weight, meaning you can fire 1 handed just not akimbo.
- the .50 Desert eagle has become a medium weight, the other deagles are still light. .50 vs .44 moment.
- AMR, auto sniper, and auto-shotty have become heavy weight, meaning you need both hands to shoot them.
- Double barrel shotgun now has unique sawing down feature making the weapon medium weight instead of heavy once sawn.

Doulbe-tested to make sure the sawn off feature works, which it seems to work just fine. Double barrel's fun, yay!!!

## Why It's Good For The Game

Was doing testing and noticed nearly all weapons, including the auto-sniper, could be akimboed. This has some problems for certain guns, making them a bullet-hell to deal with or even locking up my game when it came to Uzis as they began throwing errors eventually.

When mentioning it to Xeon shortly ago they said make all weapon weight for SMGs and AR's medium, which I fully agree with, given the bullet-hell of dual wielding AKs or SMG variants was hell on earth.

Only one I was 'meh' on was the .50 deagle but, given it's really strong for a pistol, think it's fair to medium weight it. It can be ordered via cargo console anyway so.. not like it's unqiue.

## Changelog

:cl:
balance: Added in accurate weapon weight for all darkpack guns. (SMGs and AR's = medium, snipers/autoshotty = heavy)
/:cl:


chazzyjazzy (approved):
ok - only part thats iffy is the sawoff override.

PR #897, opened 2026-04-10
Fixes Drink Dry/Diablerie Issue + Adds Prompt

Stutternov:
## About The Pull Request

Small oversight in the drink dry proc which would check if the victim was a Kindred but the source was not to try and do diablerie; properly adjusted to instead be checking if both are kindred. Otherwise it was going to the else section and instantly final-deathing any kindred that was drunk dry and unable to do diablerie.

Also, with diablerie, added a prompt to it as suggested.
image

This way you won't accidently commit diablerie without intending to. This may want a 'do-after' time, as it's still instant the moment you hit 'yes', but wasn't sure on that.

## Why It's Good For The Game

Fixes a known issue with being unable to commit diablerie. 

Prompt disallows the "I didn't mean to" excuse OOCly and provides feedback to the player that they are about to do this.

## Changelog
:cl:
add: Adds new prompt 'yes/no' prompt to diablerie.
fix: Fixes check on diablerie on the drink dry proc, diablerie now able to be done + no more FDing kindred just for hitting 0 BP
/:cl:


modular_darkpack/modules/blood_drinking/code/overfeeding/diablerie/handle_diablerie.dm
```diff
@@ -1,39 +1,43 @@
 /mob/living/carbon/human/proc/handle_diablerie(mob/living/victim)
 
-	var/datum/splat/vampire/kindred/kindred = get_kindred_splat(src)
-	var/generation = get_generation()
-	var/victim_generation = victim.get_generation()
+	var/diablerie_prompt = tgui_input_list(src, "Attempt to diablerize [victim]?", "Diablerize", list("Yes", "No"), "No")
```
FalloutFalcon: `tgui_alert` would work better here.

FalloutFalcon (changes requested):

chazzyjazzy (approved):

modular_darkpack/modules/blood_drinking/code/overfeeding/diablerie/handle_diablerie.dm
chazzyjazzy: do this actually rebel

Stutternov: Updooted to be a tgui_alert

FalloutFalcon (approved):

PR #916, opened 2026-04-13
Removes drop pods from Supply

FalloutFalcon:
## About The Pull Request
bandaid for the express console (as the default type is what should have the old train behavoir mostly)
removes pods in favor of it magicly appearing in supply (with a preferance for an open pallet) with the flavor that it was just "already in stock"

https://github.com/user-attachments/assets/76721401-bb1d-43f2-9c0a-383a53a84f38

qol so you can see through these stupid train carts
https://github.com/user-attachments/assets/d13235ad-6c14-46ba-9475-f3a1d87d452c

## Why It's Good For The Game
Lets Supply fufill the fantasy of being a well stocked and shady black market. Im prob gonna reimplement the train but i want to do a pretty major rework and this sounded way easier.
## Changelog
:cl:
fix: Cargo deliveries now appear "already in stock" somewhere in supply, with a strong preference for appearing on-top of a pallet
qol: Supply yard cargo trains are see-through when behind
/:cl:


code/modules/cargo/expressconsole.dm
```diff
@@ -270,16 +280,32 @@
...
 				new /obj/effect/pod_landingzone(landing_turf, pod_type, order)
+			*/
+			// DARKPACK EDIT ADD START
+			var/crate
+			if(order.pack.crate_type)
+				crate = order.generate(landing_turf)
+			else if(order.pack.order_flags & ORDER_GOODY) //Goody orders lack a crate_type and need special handling
+				crate = order.generateCombo(landing_turf, order.orderer, order.pack.contains, order.pack.cost)
+
+			to_chat(user, span_notice("[crate] is definitly lying around [landingzone] somewhere in the stock."))
```
chazzyjazzy:
```suggestion
			to_chat(user, span_notice("[crate] is definitely lying around [landingzone] somewhere in the stock."))
```

chazzyjazzy (approved):
dont really get the see_through_cargo_crate stuff but otherwise seems fine and most people who play supply say yes.

obviously i think in the future having some better system like a cargo ship arriving would be better.

PR #922, opened 2026-04-14
Changes Exploitable Records text to match Eyes of Chaos

Biplume:
## About The Pull Request

Changed the exploitable records default text to more accurately describe how Eyes of Chaos reveals information.

image

## Why It's Good For The Game

A lot of people on TFN seemed confused by Eyes of Chaos so I figured this would at least help clear it up a bit.

## Changelog
:cl:
spellcheck: Exploitable records description now more accurately describes the intended contents.
/:cl:

### DIFF
--- a/code/__DEFINES/~darkpack/flavor_text.dm
+++ b/code/__DEFINES/~darkpack/flavor_text.dm
@@ -14,4 +14,4 @@
 /// How many characters will be displayed in the flavor text preview before we cut it off?
 #define FLAVOR_PREVIEW_LIMIT 110
 /// The default value that will go in any new player's exploitables.
-#define EXPLOITABLE_DEFAULT_TEXT "This is where you put flaws that can be exploited in any way. This will be viewable by certain discplines if you modify this string, but only if there's anything at all in this box."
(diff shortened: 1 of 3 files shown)


FalloutFalcon:
This does imply we ONLY ever use it for Eyes of Chaos. If we are gonna do that, we might want to rename it to be more clear about that or something. I think the reason it was worded this way originally is because the intent was we might use it else where.

"Records" mostly is a hold over from it being ported from ss13 servers were its something an antag is able to read and would maybe be written in the same way you would medical or sec records.

dwinters99:
I agree with Fallcon in that leaving it generic-ish so we can use it for other things might be helpful

Biplume:
I mean, the single usage of it now isn't meant to be literal information. What should I do for that usage of it, then?

chazzyjazzy (changes requested):
as this pr exists right now, as abby correctly noted, this implies exploitable records is something that will only ever exist to serve eyes of chaos - which is not the case. don't think any updated wording for exploitable records is really necessary. not sure what to do about eyes of chaos and the associated 'these things should be imagery or whatever and not literal' as people are having the same problem with other abilities even in spite of very very clear wording (auspex 4 telepathy get_flavor_text proc and the associated confusion on discord of people saying 'its not clear how im supposed to answer'...)

so im close keying, dont think anything needs to be done regarding the fundamental premise of this pr

edit: we dont have close key flags so just take it as you will

PR #933, opened 2026-04-16
Canon Name for Strip Club in Phones

Knightscholar:
## About The Pull Request

Removes reference to any canon name for the Primogen Toreador
## Why It's Good For The Game

Easier to use whichever name any given server uses.
## Changelog
:cl:
remove: Removed reference to "Rosebud" in the Primogen Toreador phone contact
/:cl:


chazzyjazzy (approved):

dwinters99:
Probably better to just strip the name from it all together considering not all the downstream maps have the same club name

FalloutFalcon (approved):
The sprites/signage also reflect this. IDK where this name came from.

Knightscholar:
> Probably better to just strip the name from it all together considering not all the downstream maps have the same club name

Very fair suggestion. Split the difference by just removing reference to the club name and keeping it "Night Club"

chazzyjazzy:
if the signs say dora, the phone should say dora. can't we use a define for this?

FalloutFalcon (changes requested):
define is a good idea. we already have a folder full of brands that was mostly designed to solve this issue.

`code/__DEFINES/~darkpack/branding.dm`

chazzyjazzy:
is now an empty pr thanks to #951

PR #939, opened 2026-04-18
joining with negative freebie points is not allowed

chazzyjazzy:
## About The Pull Request

adds a tgui alert if you try to join the game with negative freebie points

dreamseeker_lPhU0yQMZZ
kvJ9k1mJpb

## Why It's Good For The Game

preventing exploits

## Changelog
:cl:

fix: you can no longer join with negative freebie points

/:cl:

### DIFF
--- a/code/modules/mob/dead/new_player/latejoin_menu.dm
+++ b/code/modules/mob/dead/new_player/latejoin_menu.dm
@@ -166,6 +166,12 @@ GLOBAL_DATUM_INIT(latejoin_menu, /datum/latejoin_menu, new)
 					tgui_alert(owner, "The server is full!", "Oh No!")
 					return TRUE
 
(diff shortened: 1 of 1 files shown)


dwinters99:
We had some reports that quirks having their price changed in a PR recently caused people to have extra albeit non-negative freebie points. Does this catch that?

chazzyjazzy:
> We had some reports that quirks having their price changed in a PR recently caused people to have extra albeit non-negative freebie points. Does this catch that?

sadly no and the litany of exploits related to merits and flaws are something that needs to be addressed, might as well add that to the list

the main concern is the infinite freebie points exploit which comes from swapping slots (which does not refresh the merits and flaws page, for some reason, it only refreshes when you add or remove a quirk, but if u join the game, u have the appropriate quirks, thanks /tg/)

FalloutFalcon (approved):

PR #957, opened 2026-04-23
Fixing Monstrous Quirk

chazzyjazzy:
## About The Pull Request

negative st_add_stat_mods do not work

## Why It's Good For The Game

negative st_add_stat_mods now work

## Changelog

:cl:

fix: negative stat modifiers now work

/:cl:


FalloutFalcon (approved):
a more generic ways to clamp stats so that stuff like this is more resilient to the effects of other buffs might be desirable in future.

PR #961, opened 2026-04-25
invisible-izes the ugly missing belt.dmi iconstate

chazzyjazzy:
## About The Pull Request

the purple counter strike source missing texture thing must go

## Why It's Good For The Game

its ugly

## Changelog

:cl:

del: deletes the belt missing icon state

/:cl:


FalloutFalcon (changes requested):
non modular edit but its very intentional for this thing to exist. i means someone needs to actually give an item an icon state so it can be seen.

chazzyjazzy:
> non modular edit but its very intentional for this thing to exist. i means someone needs to actually give an item an icon state so it can be seen.

im aware why it exists but people just dont use the slot in game because of it and since nearly all of our items dont have belt slot spriets the slot is just completely unused by most people

FalloutFalcon:
It you really want to have items be invisible, i think its better to declare it PER item, by setting the worn icon state to null. that way there is more of a clear delineation on code that a type path is missing a asset rather then someone fucking up a icon-state accidentally, where it would therefore clearly display in game that something is wrong.

chazzyjazzy:
> It you really want to have items be invisible, i think its better to declare it PER item, by setting the worn icon state to null. that way there is more of a clear delineation on code that a type path is missing a asset rather then someone fucking up a icon-state accidentally, where it would therefore clearly display in game that something is wrong.

theres alot of items to cover tbh because afaik we have like no belt icon sprites for any vtm codebase objects

what comes to mind immediately are keys, certain guns, etc

dwinters99 (changes requested):
it exists for a reason. either make belt sprites or make a good looking 'generic' sprite.

chazzyjazzy:
lol.

PR #962, opened 2026-04-25
Street Sign that shows area name

FalloutFalcon:
## About The Pull Request
a subtype for the street sign icon_state that displays the area its in.
image

A subtype of for all the other street signs that we had in the dmi.

## Why It's Good For The Game
Where am I?
## Changelog
:cl:
qol: Street signs tell you where you are
/:cl:

### DIFF
--- a/_maps/map_files/Vampire/runtimetown.dmm
+++ b/_maps/map_files/Vampire/runtimetown.dmm
@@ -3051,6 +3051,11 @@
 "Zz" = (
 /turf/open/misc/grass/random/grass,
 /area/vtm/outside/pacificheights)
+"ZC" = (
+/obj/effect/turf_decal/bordur,
+/obj/structure/roadsign/street,
+/turf/open/floor/plating/sidewalk/poor,
+/area/vtm/outside/pacificheights)
 "ZF" = (
 /obj/structure/railing,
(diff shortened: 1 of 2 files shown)


chazzyjazzy (approved):

Marshmellow105:
YES !!

dwinters99 (changes requested):
this is extraordinary but needs a var or subtype that allows doing it manually. not all area names are suitable for this and making an actual streetname isn't possible with this system.

FalloutFalcon:
> this is extraordinary but needs a var or subtype that allows doing it manually. not all area names are suitable for this and making an actual streetname isn't possible with this system.

sure. notibly we do still have a few assets in here that dont have subtypes so require varedits anyway.. mabye i should give them those.

FalloutFalcon:
> this is extraordinary but needs a var or subtype that allows doing it manually. not all area names are suitable for this and making an actual streetname isn't possible with this system.

also fyi, this is so people can locate an area if the game yells at them to go to it (police radio or masq contract and stuff) so if we cant reliably use the area name, thats prob its own issue.

FalloutFalcon:
just a bit of type path bloat. 
image

dwinters99 (approved):

PR #948, opened 2026-04-22
tfn upstream - ghouls and harddels

chazzyjazzy:
## About The Pull Request

makes ghouls only able to take level 1 in any given discipline
makes ghouls get potence 1 for free

solves ghoul splat harddels by removing bloodheal and potence on_lose_or_destroy

fixes a bug caused by our flawed implementation of destroy procs that i didnt update the pr with

## Why It's Good For The Game

fix fix fix fix

## Changelog
:cl: buffyuwu
add: ghouls now get potence 1 for free
fix: ghouls can now only have level 1 in their disciplines
fix: solves bloodheal and potence harddels by removing the disciplines on_lose_or_destroy
fix: fixes a bug with transformation disciplines causing disciplines to become unusable
/:cl:

Issue #869, opened 2026-04-05
all transforms forbid the use of disciplines


chazzyjazzy:
### Issue Summary:

all transformations now remove your ability to use disciplines forever.

<img width="823" height="618" alt="Image" src="https://github.com/user-attachments/assets/1bae2355-b0ff-4b84-801c-60daf065953f" />

### Reproduction Steps:

1. transform
2. duuude
3. i cant use disciplines

Dangerlurking:
This did not happen to me when I transformed into Zulo as a Tzimisce, but Zulo form itself has a lot of issues rn.

chazzyjazzy:
> Is this related to us trying to transfer the discs to the mob but then deleting them after it gets deleted.

no it was because of the harddeletes fix qdeleting 'discipline' on action/discipline and on discipline_power when in reality neither shouldnt have been doing that because datum/discipline was the holder for the other two rather than action holding discipline and discipline holding power

fixed in #948 which im gonna merge in like the next hour

PR #976, opened 2026-04-27
disciplines should have a destroy() too

buffyuwu:
## About The Pull Request

adds a destroy() for disciplines to hopefully fix harddels

### DIFF
--- a/modular_darkpack/modules/powers/code/discipline_actions.dm
+++ b/modular_darkpack/modules/powers/code/discipline_actions.dm
@@ -52,6 +52,10 @@
 	QDEL_NULL(discipline)
 	return ..()
 
+/datum/action/discipline/Destroy()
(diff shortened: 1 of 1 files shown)


chazzyjazzy (changes requested):
This is actually the cause of #869 

so this shouldnt be merged most likely

FalloutFalcon:
> This is actually the cause of #869
> 
> so this shouldnt be merged most likely

In theory the lines above it cause the same issue. The action should not be able to delete the disc datum if we want to transfer discs to transformations.

chazzyjazzy:
> > This is actually the cause of #869
> > so this shouldnt be merged most likely
> 
> In theory the lines above it cause the same issue. The action should not be able to delete the disc datum if we want to transfer discs to transformations.

thats why that line is removed in #948

PR #977, opened 2026-04-27
Removes dice output preference

FalloutFalcon:
## About The Pull Request

Removes Dice output preference in favor of just always outputting to chat unless its marked spammy per roll
## Why It's Good For The Game
This made alot more sense when they were more equivlent, pretty much just showing the success.
The chat output has slowly outpaced balloon alert in terms of information (now including stuff like the bumper text and feedback on what dice were used). This is now a noob trab which just causes alot of confusion because they dont know what the roll is acctually for, and there was a point in time where it was the default which creates issue for people who dont know its a setting.
## Changelog
:cl:
del: Removes dice output preference in favor of always outputting to chat unless declared by the roll.
/:cl:


chazzyjazzy (approved):
yea, the chat box just gives so much more info, and so much more useful info, than the balloons

PR #983, opened 2026-04-29
Config for roleplay only merits

FalloutFalcon:
## About The Pull Request
Config for if role-play only quirks/merits are loaded and available.

Saw https://github.com/The-Final-Nights/The-Final-Nights-Rebase/pull/245 and realized this should have existed for this exact purpose.

This does not include stuff that's hooks for roleplay but is still clearly conveyed to other players (like permenent fangs) as you cant just pretend they dont exist and can have real ramifications for some of them.

## Why It's Good For The Game
Some servers dont want things  that act as "freebie" points that you dont "have" to roleplay in theory.
## Changelog
:cl:
config: Adds config for if roleplay only merits are available
/:cl:


chazzyjazzy (approved):

chazzyjazzy:
what about for the werewolf merit

FalloutFalcon:
> what about for the werewolf merit

Not aware of any other merits without mechanical effect to some degree?

FalloutFalcon:
Skimming the list, I assume you meant `/datum/quirk/darkpack/animal_musk`, which does still show up as flavor text, the same as permenent fangs, so unless we also want a seperate config for flavor text only stuff it doesnt fit in this pr. (I did go update the desc to remove the lines that arent mechanically implemented yet)

PR #984, opened 2026-04-29
Removes unticked and unplanned delivery module

dwinters99:
## About The Pull Request
title
## Why It's Good For The Gamea
we're nearing a post-ticked file enforcement world and this stuff isn't used and isn't planned for use. Dispose of it At Once My Liege.
## Changelog

No player-facing changes


FalloutFalcon (approved):

chazzyjazzy (approved):
yea that makes sense. it would be nice to bring something like this back in full force but lets be so real nobody likes this minigame or uses it...

FalloutFalcon:
> yea that makes sense. it would be nice to bring something like this back in full force but lets be so real nobody likes this minigame or uses it...

It didnt exist on tfn or apoc so i have no clue how liked it acctually was on wod13

PR #986, opened 2026-04-29
Adding factions to jobs

chazzyjazzy:
## About The Pull Request

adding factions to jobs, these are needed for gameplay loops that are planned, they are currently all 'FACTION_CITY' because for some reason in jobcode if jobs arent FACTION_STATION (renamed to FACTION_CITY in our modular edits) then you just cant select them in prefs. 

## Why It's Good For The Game

i can use this var for gameplay loops

## Changelog

:cl:

code: every job datum now has a faction
code: removes unused and useless 'is_gun_store' job on /obj/structure/retail

/:cl:

### DIFF
--- a/code/modules/client/preferences/middleware/jobs.dm
+++ b/code/modules/client/preferences/middleware/jobs.dm
@@ -18,8 +18,10 @@
 	if (isnull(job))
 		return FALSE
 
+	/* DARKPACK EDIT REMOVAL - Factions - note: why do these lines even exist?
 	if (job.faction != FACTION_CITY) // DARKPACK EDIT, ORGINAL: if (job.faction != FACTION_STATION)
 		return FALSE
+	*/
 
 	if (!preferences.set_job_preference_level(job, level))
 		return FALSE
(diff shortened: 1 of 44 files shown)


code/modules/client/preferences/middleware/jobs.dm
```diff
@@ -18,8 +18,10 @@
 	if (isnull(job))
 		return FALSE
 
+	/* DARKPACK EDIT REMOVAL - Factions - note: why do these lines even exist?
 	if (job.faction != FACTION_CITY) // DARKPACK EDIT, ORGINAL: if (job.faction != FACTION_STATION)
 		return FALSE
+	*/
```
FalloutFalcon, line 24: This line exists because the var only exists to determine whether a job is considered crew in prefs.
Its the only usecase outside of anonymous code.

~~Im not even certain it actually sets faction in how it exists for mob targeting behavior.~~
It does not add to your actual faction list. I checked.

chazzyjazzy: ill make it so it adds - kinda unintuitive that it doesnt add to the faction list...

FalloutFalcon (approved):
mabye a slight misuse of the intent of the varible. might lead to minor issue if behavior is introduced to upstream but thats nothing.

PR #992, opened 2026-04-30
Prevents phone frequency collision

FalloutFalcon:
## About The Pull Request
Upstreaming of https://github.com/The-Final-Nights/The-Final-Nights-Rebase/pull/208 with some bonus stack trace included
## Why It's Good For The Game
The old code was weirdly dense and buggy.
## Changelog
:cl: FalloutFalcon, CYBILIZE
fix: fixed separate phone calls occasionally sharing audio with each other
code: corrected secure phone frequency assignment logic
/:cl:


chazzyjazzy (approved):
thanks for the stack trace because there was some worry about using a while loop

FalloutFalcon:
This code is simple enough that i can concretely say it was impossible for that to cause an infinite loop. It could cause lag spikes if something was generating a insane volume of entries in the list which this mostly prevents.

PR #998, opened 2026-05-01
Fix in the logic of Area power

FalloutFalcon:
## About The Pull Request
idk.


chazzyjazzy:
i found it out!!! since /tg/ uses power_(whatever) vars, and fuseboxes modify those vars, we should probably just check for one of those on lampposts, this allows us to leave the areacode unchanged as we're not interacting with powered or requires_power.

edit: and if we WANT to have an area that starts off unpowered or whatever, we can just set that in the area define since thats where /tg/ sets it anyways
dreamseeker_ZLMi0wrP2F

dreamseeker_mq0811k91f

chazzyjazzy (approved):
just gonna approve since i think we fixed it in a good way without tackling an entire rework of the spaghetti

PR #1001, opened 2026-05-02
Garou bites do agg, Garou speed nerfs

ghost:
## About The Pull Request

This nukes most of the Garou speed buffs, keeping a toned-down one for Hispo and Lupus.
It also moves the AGGRAVATED damage to a child of fera organs, as only certain forms get this. In particular, in the W20 core, Glabro and Lupus claws explicitly do not agg, but Crinos, Hispo, and Lupus bites do.

## Why It's Good For The Game

Right now, a Crinos is equivalent to Dex 8 or 9. That means that even using Celerity, you'll get caught up to. That's not exactly great for something that does a fuckton of melee damage all at once and potentially more easily soon. I was moving so fast on Lupus between my base dex speed and the +2 dex and the -0.5 that I was actually having trouble keeping track of what I was doing
## Changelog
:cl:
balance: Crinos, Hispo, and Lupus speeds were all nerfed.
balance: Hispo and Lupus bites do aggravated damage.
/:cl:

### DIFF
--- a/modular_darkpack/modules/werewolf_the_apocalypse/code/species/garou_organs.dm
+++ b/modular_darkpack/modules/werewolf_the_apocalypse/code/species/garou_organs.dm
@@ -6,13 +6,15 @@
 	head_flags = NONE
 	unarmed_attack_sound = 'modular_darkpack/modules/werewolf_the_apocalypse/sounds/werewolf_bite.ogg'
 
+/obj/item/bodypart/head/fera/aggravated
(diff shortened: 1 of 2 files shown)


FalloutFalcon (commented):
> Glabro claws don't do agg

They already didn't and this code only affects LUPUS's claw damage (which barely matters as current code forces them to bite)

> I should give the bites +1 dice
Image

FalloutFalcon (approved):

PR #1007, opened 2026-05-02
Reworks active potance effects

FalloutFalcon:
## About The Pull Request
Active potance is now mostly TTRPG accruate and its primary effect is converting potances strength dots directly into AUTO SUCCESSES (regardless of dot cast, which means all the actions now do the same thing. (investigate a flag for not adding some actions tbh)
Acctually a pretty massive buff to potance as it makes damage VERY consistent.

I was originally gonna tackle Vengeance of Samiel cause it could use some work but ATOMIZATION!!

I fixed the melee thing but its kinda due for scraping once melee is affected by strength

https://github.com/user-attachments/assets/9ea24806-ec32-44ac-bb96-ab3d41a6d617

As a refresher: V20 pg. 192
image

## Why It's Good For The Game
TTRPG accuracy. Makes brawl actually effected by potance.

## Changelog
:cl:
add: Adds support for automatic successes based on the TTRPG
code: Cleans up some of the code surrounding potance
balance: Kills some of the snowflaky effect of potance in favor of it granting automatic successes (TTRPG accuracy)
qol: Makes the dice unicode characters to show individual dice results a little larger
fix: Potance 1-2 no longer cripples melee damage and it now scales correctly with all dots.
/:cl:

### DIFF
--- a/modular_darkpack/modules/powers/code/discipline/potence/potence.dm
+++ b/modular_darkpack/modules/powers/code/discipline/potence/potence.dm
@@ -30,6 +30,9 @@
 		/datum/discipline_power/potence/five
 	)
 
+/datum/discipline_power/potence/post_gain()
+	owner.st_add_stat_mod(STAT_STRENGTH, level, "Potence")
+
 /datum/discipline_power/potence/one/activate()
 	. = ..()
 
@@ -40,8 +43,6 @@
 
 	owner.remove_status_effect(/datum/status_effect/potence/one)
 
-/datum/discipline_power/potence/one/post_gain()
-	owner.st_add_stat_mod(STAT_STRENGTH, 1, "Potence")
 
 //POTENCE 2
 /datum/discipline_power/potence/two
@@ -70,8 +71,6 @@
 	. = ..()
 	owner.remove_status_effect(/datum/status_effect/potence/two)
 
-/datum/discipline_power/potence/two/post_gain()
-	owner.st_add_stat_mod(STAT_STRENGTH, 2, "Potence")
 
 //POTENCE 3
 /datum/discipline_power/potence/three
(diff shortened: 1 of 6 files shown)


modular_darkpack/modules/storyteller_dice/code/roll_datum.dm
```diff
@@ -151,6 +162,8 @@
 				extra_dice++
 		for(var/i in 1 to extra_dice)
 			rolled_dice += rand(1, sides)
+	for(var/i in 1 to auto_successes)
+		rolled_dice += 11
```
chazzyjazzy, line 166: 11? might wanna check if this is gonna take a sledgehammer to the double successes vs exploding dice config?

chazzyjazzy (approved):
ok looks good - but may wanna doublecheck it wont wreck the exploding dice whatever logic and also that auto successes are accounted for in update_modifiers_from_stat_whatever, maybe sticking update_auto_successes in there or whatever's necessary.

u were right that this is really strong. we'll know quickly if its too much.

modular_darkpack/modules/storyteller_dice/code/roll_datum.dm
FalloutFalcon: they work off of checking 10 specifically.

FalloutFalcon:
> ok looks good - but may wanna doublecheck it wont wreck the exploding dice whatever logic and also that auto successes are accounted for in update_modifiers_from_stat_whatever, maybe sticking update_auto_successes in there or whatever's necessary.

It functionally acts as a bonus for what uses it unless you specifically flag otherwise.

modular_darkpack/modules/powers/code/discipline/potence/potence.dm
```diff
@@ -7,41 +7,49 @@
...
+	check_flags = DISC_CHECK_CAPABLE
+
+	toggled = TRUE
+	duration_length = 1 TURNS
+
+/datum/discipline_power/potence/post_gain()
+	owner.st_add_stat_mod(STAT_STRENGTH, level, "Potence")
+
+/datum/discipline_power/potence/post_loss()
+	owner.st_remove_stat_mod(STAT_STRENGTH, "Potance")
```
chazzyjazzy:
```suggestion
	owner.st_remove_stat_mod(STAT_STRENGTH, "Potence")
```

PR #1014, opened 2026-05-04
Displaying area specific stuff when entering into a new area

chazzyjazzy:
## About The Pull Request

drafted because i have a few todos
- change it from area/entered into comsig_area_entered probably on mob/living
- make a better system for this in general because ideally we want it to be easily added onto / taken away from for future additions like domains
- this whole thing just needs to be entirely on the mob/living level
- add cooldown so that you dont get spammed

## Why It's Good For The Game

cool

## Changelog

:cl:

qol: area title and info is displayed when crossing areas

/:cl:


chazzyjazzy:
not sure i like this too much... not sure why...

FalloutFalcon (approved):

PR #1015, opened 2026-05-04
Reworks Melee to use rolls and stats

FalloutFalcon:
## About The Pull Request
Mostly reproduces the same code as brawl.
This needs a followup pr with a ton of balance tweaks for the weapons, we have some varibles that lets us replicate ttrpg stuff. (like attack difficulty) 

## Why It's Good For The Game

This has alot of balance effects but TTRPG accuraecy and parody with brawl is good.
## Changelog
:cl:
balance: Reworks melee to roll damage based off two rolls (to parody the TTRPG and brawl)
balance: Protean claws use brawl
refactor: Removes snowflake protean claws damage buff in favor of generic melee code
/:cl:


code/_onclick/item_attack.dm
```diff
@@ -302,12 +302,15 @@
...
-// DARKPACK EDIT ADD START
-	if(isliving(user))
-		var/mob/living/living_user = user
-		var/stat_multiplier = living_user.st_get_stat(STAT_MELEE) * 0.4
-		final_force *= stat_multiplier
-// DARKPACK EDIT ADD END
+	// DARKPACK EDIT ADD START - STORYTELLER_ROLLS/STORYTELLER_STATS
+	// This is pretty evil, but we are gonna convert all the tg force into the +# that melee weapons have listed.
+	// This means we can do stuff like set force of a baseball bat to 2 TTRPG_DAM and it just works.
+	var/bonus_dice = round(final_force / (1 TTRPG_DAMAGE))
```
chazzyjazzy: small nit - doesn't round() always round down? i noticed this in testing for something a bit back. i can imagine people getting antsy about this

code/_onclick/item_attack.dm
```diff
@@ -302,12 +302,15 @@
...
+	var/datum/storyteller_roll/damage/damage_roll = new()
+	damage_roll.applicable_stats = list(attacking_item.st_damage_stat)
+	var/damage_roll_result = damage_roll.st_roll(user, src, bonus_dice)
+	final_force = damage_roll_result
```
chazzyjazzy: don't you mean damage_roll_result TTRPG_DAMAGE?

i forget if this is already handled by the storyteller roll subtype

code/_onclick/item_attack.dm
```diff
@@ -348,8 +351,24 @@
...
 
-	// DARKPACK EDIT ADD START - (Makes Melee do Something)
-	final_force += final_force * (user.st_get_stat(STAT_MELEE) * 0.1)
+	// DARKPACK EDIT ADD START - STORYTELLER_ROLLS/STORYTELLER_STATS
+	var/datum/storyteller_roll/attack/attack_roll = new()
+	attack_roll.applicable_stats = list(attacking_item.st_attack_ability, attacking_item.st_attack_attribute)
+	attack_roll.difficulty = attacking_item.attack_difficulty
+	var/attack_roll_result = attack_roll.st_roll(user, src)
+
+	// What i want to do is acctually have it return if not success. But that creates bad visual feedback as all the FX still play. just give them them SOME damage..
```
chazzyjazzy: good idea since the rng-ification of combat would probably be ferociously hated

code/_onclick/item_attack.dm
```diff
@@ -348,8 +351,24 @@
...
+	if(attack_roll_result == ROLL_SUCCESS)
+		// This is pretty evil, but we are gonna convert all the tg force into the +# that melee weapons have listed.
+		// This means we can do stuff like set force of a baseball bat to 2 TTRPG_DAM and it just works.
+		var/bonus_dice = round(final_force / (1 TTRPG_DAMAGE))
+		var/datum/storyteller_roll/damage/damage_roll = new()
+		damage_roll.applicable_stats = list(attacking_item.st_damage_stat)
+		var/damage_roll_result = damage_roll.st_roll(user, src, bonus_dice)
+
+		final_force = max(damage_roll_result TTRPG_DAMAGE, 1 TTRPG_DAMAGE)
```
chazzyjazzy: so i don't get it - say a baseball bat has original force = 40. this makes bonus dice = 4. so if i have 0 strength, then the roll becomes 4 dice rolled for damage, giving me 10-40 damage which seems good. but then if i have 4 strength, then i have 8 potential dice, no? meaning strength, the stat which affects the damage roll, is actually additive to the force, not the foundation of the damage, as now my range is 10-80.

what i'm trying to say here is that this (the usage of ttrpg accurate rolls in melee) is an additive buff to the calculation of melee force, rather than a reworking of how the damage is created in the first place (which may be the intent of the pr?)

i don't exactly have an issue with this moreso than i feel like it may unintentionally up-end the balance - i see what you were trying to do, and it is smart

then again if the point is to have a better bandaid than my 'makes melee do something' pr, then this definitely does that, i think, if a bit stronger

chazzyjazzy (changes requested):

code/_onclick/item_attack.dm
FalloutFalcon: I mean brawl has the same rng functionally (actually worse since there is less bonuses you can get for brawl) and its kinda fine.

code/_onclick/item_attack.dm
FalloutFalcon: my idea is so that i can use the old force varible without to much hastle and still get mostly accurate numbers. This lets me use force as a representation of the + # that weapons have. 

VTM and WTA. ~~Oh god these charts aren't the same. I think i will use wta numbers cause its more complete. We will get to that when we get to that.~~
Image

Image

So if i set the weilded axe damage to `force = 3 TTRPG_DAMAGE` it comes out to mostly ttrpg accurate. That works out to a small buff for the axe even. Skimming the numbers most weapons prob wont change too much. Most of the significate stuff also has lower difficutly (knife damage drops a ton but it has lower difficulty)

(as in the ttrpg) it means that the bonuses of weapons scales alot harder when you have shitty stats but high stat characters can kinda john-wick most weapons. Its interesting for sure.

code/_onclick/item_attack.dm
FalloutFalcon: Yea. That should only really matter for things that are less then 1 TTRPG_DAMAGE or didnt have a balance pass after this pr. (or weird stuff that manipulates force values like the sharpening stone)

code/_onclick/item_attack.dm
FalloutFalcon: nope. seems like a mistake on my part.

code/_onclick/item_attack.dm
chazzyjazzy: it is interesting - and what youre doing is smart even without an audit of damage numbers or whatever, but again i feel like this would make melee weapons far stronger on characters who are built for it, not that i have a problem with that per se, i just feel like perhaps that was not intended?

FalloutFalcon:
wait this hella fails unit tests atm. I thought i had already fixed that.

code/_onclick/item_attack.dm
FalloutFalcon: its kinda just how it works out in the ttrpg, it makes sense if you invest heavy to be rewarded..

chazzyjazzy (approved):
it looks like my review was mostly addressed so heres an approve, but i will use my crystal ball for a moment and predict that longswords which are already quite strong will become even stronger so prepare for that

PR #1020, opened 2026-05-05
Delirium

FalloutFalcon:
## About The Pull Request
Reallllly similar to the work I did for frenzy. 
A role-play driven status effect that shows some cool visuals and messages to tell you what your character should be doing rn.

https://github.com/user-attachments/assets/608d1f25-a37c-4a44-8ca0-d0a690eadd09

## Why It's Good For The Game
Neat pointer that there is infact a evil monster that previously fucked up your species so bad you have generational magical trauma.
## Changelog
:cl:
add: Delirium status effect
/:cl:


FalloutFalcon:
I bashed my head against a wall trying to get these to render direction correct but I might just raw not be able to via `.images` (lummox said use vis contents which i cant because it needs to render to 1 person)
image

chazzyjazzy (approved):

PR #1049, opened 2026-05-11
many more private rolls

ghost:
## About The Pull Request

Made a lot of rolls private, some now PRIVATE_AND_TARGET as well.
## Why It's Good For The Game

Many discs and gifts should be private. For the Fianna lights for example, it literally says it can be used for ambushes. Blur of the Milky Eye is a stealth gift. Frenzy rolls essentially became a reminder of "THIS GUYS A VAMPIRE" every few seconds even if succeeded. And Samedi discs could kill you without saying why. And so on.


FalloutFalcon:
frezy rolls are the only one im a little iffy on as they create alot of visible drama which i think is neat.
I still kinda feel as if a "party" system would be cool so you can display rolls to people you want them to be displayed to.

FalloutFalcon (approved):

PR #1050, opened 2026-05-13
Firearms Stat Version 1.1 (Patched, Updated, Re-Released)

Stutternov:
## About The Pull Request

Basically does the following:
- Firearms stat **no longer** has a probability chance of doing 40% extra damage  (used to be firearm skill x 10 as prob to do 40% bonus damage)
- Firearm roll that existed already for recoil has gotten a face-lift. It now includes a reduction to spread based on the roll you did. The difficulty of the roll is now set to 6, adding +1 difficulty for akimbo or burst firearms and +2 for automatic weapons.
- Firearms have had mild tweaks to recoil and spread to accommodate for this. 

Seen here firing a bolt-action, a burst, and a full-auto
image

## Why It's Good For The Game

Overall this brings it a bit more in line with tabletop. 

It was noticed on TFN that if you rolled that 40% bonus 'crit' it could make calibers bypass armor due to raw damage output. This wasn't a huge thing but it was asked 'why is this the case' - and realized in TT firearms skill works differently.

I didn't replicate the accuracy factor on TT since automatics do not work how they do on tabletop but attempted to simulate the method in which you roll for spread and recoil instead.

This should make guns require dots to properly use them well rather than just random bonuses to damage. Felt like using full autos were strong but kicked bad, encouraging me to use burst fire guns or semis the moment I had lower skills. Even failing a 9mm roll doesn't feel that bad due to low spread low recoil but failing a roll on an AK, and Aug, or a shotgun becomes very notable. 
## Changelog

:cl:
add: Adds recoil calc stat check to reduce spread as well
add: Adds difficulty to recoil calc
del: Removes bonus damage random chance for firearm damage based on skill probability check
balance: Redoes some weapons recoil and spread slightly; overall changes ranged from 1-2 recoil/spread so very minor.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/ballistic.dm
+++ b/code/modules/projectiles/guns/ballistic.dm
@@ -613,8 +613,20 @@
 		recoil_roll = new()
 
 	if(COOLDOWN_FINISHED(src, recoil_skill_check))
+		for(var/obj/item/gun/gun in user.held_items)
+			if(gun == src || gun.weapon_weight >= WEAPON_MEDIUM)
+				continue
+			else if(can_trigger_gun(user, akimbo_usage = TRUE))
+				recoil_roll.difficulty += 1	//Akimboing gains +1 difficulty
+		if(firing_burst == TRUE)
+			recoil_roll.difficulty += 1	//Bursts gains +1 difficulty
+		var/datum/component/automatic_fire/automatic_check = src.GetComponent(/datum/component/automatic_fire)
+		if(automatic_check)
+			recoil_roll.difficulty += 2	//Full-auto gains +2 difficulty
+
 		var/recoil_reduction = recoil_roll.st_roll(user, src)
 		recoil = max(initial(recoil) - recoil_reduction, 0)
+		spread = max(initial(spread) - recoil_reduction, 0)
 		COOLDOWN_START(src, recoil_skill_check, 1 SCENES)
 	// DARKPACK EDIT ADD END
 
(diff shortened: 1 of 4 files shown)


modular_darkpack/modules/storyteller_dice/code/roll_subtypes.dm
```diff
@@ -78,6 +78,7 @@
 	applicable_stats = list(STAT_DEXTERITY, STAT_FIREARMS)
 	reroll_cooldown = 1 TURNS
 	numerical = TRUE
+	difficulty = 6
```
FalloutFalcon:
```suggestion
```

6 is the default value. 
OBJECT INHERITANCE!!!!!

Image

code/modules/projectiles/guns/ballistic.dm
```diff
@@ -613,8 +613,20 @@
 		recoil_roll = new()
 
 	if(COOLDOWN_FINISHED(src, recoil_skill_check))
+		for(var/obj/item/gun/gun in user.held_items)
+			if(gun == src || gun.weapon_weight >= WEAPON_MEDIUM)
+				continue
+			else if(can_trigger_gun(user, akimbo_usage = TRUE))
+				recoil_roll.difficulty += 1	//Akimboing gains +1 difficulty
```
FalloutFalcon, line 620: suprised we dont have a more generic helper for checking akimboing tbh?

FalloutFalcon (approved):

code/modules/projectiles/guns/ballistic.dm
chazzyjazzy: yea thats really ugly but im curious as to how invasive a more clean check would be esp considering /tg/ code.

chazzyjazzy (approved):
OK it looks fine but don't think i don't notice the stealth buff to the AK and others

PR #1053, opened 2026-05-14
Cars can travel through transfer points

FalloutFalcon:
## About The Pull Request
Cleans up a little logic and makes it so A: They cant infinitely loop through transfer points via bumping into them on every car tick. B: Prevents them from getting places INSIDE of a transfer point, meaning they constantly bump it.

## Why It's Good For The Game
i hate cars. there code is so jank and creates hella weird issues.

## Changelog

:cl:
fix: Fixes interactions surrounding cars and transfer points
/:cl:


chazzyjazzy (approved):
looks fine

PR #1058, opened 2026-05-16
IrisStation Fancy Paperwork Port

Kolibri-B:
## About The Pull Request

A simple **_port_** of the following PR: https://github.com/IrisSS13/IrisStation/pull/522

This allows for paperwork to be more customizable, using a subset of allowed CSS properties with a lot of extra text sanitization.

I don't know if the maintainers want this or not, but I'm putting the PR up so they can decide.

## Why It's Good For The Game

Paperwork enjoyers can be paperwork havers!

## Changelog
:cl:
code: Expanded the set of allowed HTML tags, allowing for richer formatting.
code: Updated which HTML attributes are allowed or forbidden for improved security and flexibility.
code: Added strict CSS validation to prevent style-based exploits. Only a safe subset of CSS properties and values are now permitted.
fix: Fixed template not updating in real time when typing.
qol: Increased character limit for papers.
admin: Added logs for restricted HTML/CSS usage.
/:cl:


tgui/packages/tgui/sanitize.ts
```diff
@@ -43,13 +63,320 @@ const defTag = [
...
+  /max-width\s*:\s*\d{4,}px/i, // Very wide maximums
+  /max-height\s*:\s*\d{4,}px/i, // Very tall maximums
+];
+
+// Trusted domains for background images (empty by default for maximum security)
+const trustedDomains = [
+  // Add trusted domains here if needed
+  // 'trusted-domain.com',
+  // 'cdn.example.com',
+];
```
FalloutFalcon, line 221: Im not sure why this one exists if its "emtpy by default" but never populated by anything. Consider adding all the domains we allow for headshots here. (this might even be worth putting into a global and then passing in the static ui data)

code/modules/paperwork/paper.dm
```diff
@@ -699,11 +711,9 @@
...
 		if("add_text")
 			var/paper_input = params["text"]
-			var/this_input_length = length_char(paper_input)
-
-			if(this_input_length == 0)
-				to_chat(user, pick("Writing block strikes again!", "You forgot to write anything!"))
-				return TRUE
+			var/blocked_summary = params["blocked_summary"]
+			if(blocked_summary && blocked_summary != "")
+				log_admin("[key_name(user)] had forbidden HTML/CSS sanitized from paper: [blocked_summary]")
```
FalloutFalcon, line 719: I'm actually now wondering how easy it is to inject into current code but a notable issue here is all the validation here is being done in tgui, which due to being separate windows and tgui being quite easy to manipulate in the first place (plus some other stuff) acts as a really easy attack route.

These checks only stops the simplest of bad actors (most likely not even bad actors, just people trying to put in a tag we dont support) unfortunately. However your PR doesnt acctually the code any LESS secure, this is just an existing vurlanbitly as far as I can tell.

Kolibri-B:
Will add modularization comments soon. I thought I already did that, but I guess I forgot.

FalloutFalcon (commented):
~~None of your changes are modular. Please read `modular_darkpack/readme.md`
this will be an annoying one for it unfortunatly as it has alot of tgui edits.~~

code/modules/paperwork/paper.dm
FalloutFalcon: I acctually rubber-ducked the problem with a TG maint, I def am misunderstanding this a little. While its true that we CAN definitely get an arbitrary set of bad data in through this, and this log would almost never be sent by a true bad actor, we still SHOULD be sanitizing the data TGUI side before any code would have a chance to properly run. (assuming all saftey checks are functioning and there is no way to bypass them (which is my new minor worry)

FalloutFalcon:
Im also gonna parrot similar concerns as the TG devs gave on https://github.com/tgstation/tgstation/pull/95665
image

Kolibri-B:
> Im also gonna parrot similar concerns as the TG devs gave on [tgstation/tgstation#95665](https://github.com/tgstation/tgstation/pull/95665)

I'm not a cybersecurity expert or anything, so I make no guarantees. Some people on TFN wanted it ported over, so I just went ahead and made a PR to port it over; I'd fully expect it to be thoroughly reviewed by maintainers prior to merging. I did some research & basic exploitation testing myself and found nothing unusual, but again, not an expert.

PR #1061, opened 2026-05-18
(Draft) Potence Active Alternative - BP Consumed on Hit/Action

Stutternov:
## About The Pull Request

This is a draft currently purely for feedback. Members of TFN's community gave feedback that potence's drain at 1 turn length made it a massive BP drain for minimal returns compared to Celerity, or the fact you were draining usually 2 BP before you'd even get a roll off to bash down a door.

This was a solution floated in OOC chat and had some support seemingly in discussions. Potence consuming BP per-action rather than turn. This way punching a door will take 1 BP for the door, not a timer - or you can use potence in an active fight and only lose BP when you are attacking verses when you are unable to even swing a the mob.

**This is drafted for more open feedback**, happy to hear input or alternative ideas - or if people think it's better as is rather than this.

Also - door hitting isn't fully accounted for. It.. seems to work? But it need to def be cleaner, it's not working every time I noticed. Issue on my end with the signals I'm sure.

## Why It's Good For The Game

See above; basically listed it there. 

Tl;dr - makes potence drain your BP only when you take actions rather than at rapid rates when you typically don't even get to hit something per-turn or when actions take longer than a normal turns length.

## Changelog
:cl:
balance: Removes Potence's toggle cost but instead makes it drain 1 BP per action.
/:cl:

### DIFF
--- a/modular_darkpack/modules/powers/code/discipline/potence/potence.dm
+++ b/modular_darkpack/modules/powers/code/discipline/potence/potence.dm
@@ -1,3 +1,5 @@
+#define COMBAT_COOLDOWN_LENGTH 15 SECONDS
+
 /datum/discipline/potence
 	name = "Potence"
 	desc = "Boosts melee and unarmed damage."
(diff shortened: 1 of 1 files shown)


Stutternov:
God a suggestion to just do a change for downstream instead and that this might be even more punishing. Opting for that (for now)

PR #1064, opened 2026-05-18
Area name fixes

FalloutFalcon:
## About The Pull Request
Areas wont show the same name twice (so reentering and area or one with the same name) to not spam
fixes the mapgen areas to have real names and cleans up some varibles.
adds a flag to just not show the name in a popup incase its a weird area/name
## Why It's Good For The Game
fix
## Changelog
:cl:
fix: The same area name wont be shown twice when entering areas
fix: Generate forest has a correct area name
/:cl:


chazzyjazzy (approved):
seems good

PR #1065, opened 2026-05-18
Garou tongues based on form

ghost:
## About The Pull Request
no english in crinos/hispo/lupus
no garou tongue outside crinos
no primal tongue outside hispo/lupus (lupus could not speak primal tongue before this actually)
wasnt sure if upstream would want this but im currently taking an "upstream by default, downstream if unwanted" stance

## Why It's Good For The Game

no way to represent the fact that the tongues are harder to speak in forms other than the ones listed; partial speech isnt a thing to my understanding, just partial understanding
also its hellish to be a kinfolk and even the homids are all speaking garou
but giving the tongues to kinfolk seems like the wrong move given cubs don't even get taught the high tongue; its pretty prestigious and weighty

## Changelog

:cl:
balance: Crinos may now only speak Garou Tongue, Hispo and Lupus may now only speak Primal Tongue
balance: Crinos, Hispo, and Lupus now get minor night vision and keen noses.
/:cl:


chazzyjazzy:
i mean do i even review

PR #1079, opened 2026-05-21
Attack roll balloon alerts made more clear

FalloutFalcon:
## About The Pull Request

Adds support for a prefix to the alert plus a buffer to the alert. This only works because we know there will always be a roll on the same frame as the damage meaning we can blindly buffer it for a moment. It feels extremely hacky.

We really ought to have a system that buffers them more sanely but that would have knock on effects for other balloon alerts which worries me.

I might have a flag for putting SPECIFIC rolls into a queue. (so attack and damage) so that only they are affected by the queue? This works for now.

https://github.com/user-attachments/assets/0d2c2cb5-94cf-4381-afce-0537114019f9

## Why It's Good For The Game
Old balloon alerts were useless as the stacked messages made it impossible to parse what the results were.

## Changelog

:cl:
qol: Combat dice-rolls are displayed more clearly
/:cl:

### DIFF
--- a/modular_darkpack/modules/storyteller_dice/code/roll_datum.dm
+++ b/modular_darkpack/modules/storyteller_dice/code/roll_datum.dm
@@ -11,6 +11,9 @@
 	var/roll_output_type = ROLL_PUBLIC
 	/// This is a roll that can proc multiple times in rapid sucession and thus has weaker or less notible outputs (forced runechat and quieter dice rolls)
 	var/spammy_roll = FALSE
+	/// If set, a character or unicode appended to the front of balloon alerts to help convey what the roll is for.
+	var/alert_prefix
+	var/alert_delay
 
 	/// A lazy list of times indexed by a weakref to a mob
 	var/list/mobs_last_rolled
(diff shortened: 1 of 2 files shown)


modular_darkpack/modules/storyteller_dice/code/roll_datum.dm
```diff
@@ -81,16 +84,27 @@
...
-				roller.balloon_alert(player_mob, "<span style='color: #ff0000;'>[last_sucess_amount]</span>", TRUE)
+				create_balloon_alert(roller, player_mob, last_sucess_amount)
+
 
 	LAZYADDASSOC(mobs_last_rolled, WEAKREF(roller), list(world.time, output))
 
 	SEND_SIGNAL(roller, COMSIG_LIVING_DICE_ROLLED, src, output)
 	return output
 
+/datum/storyteller_roll/proc/create_balloon_alert(mob/living/roller, mob/player_mob, number)
```
chazzyjazzy, line 100: approved but what exactly is the difference between roller and player_mob

chazzyjazzy (approved):

modular_darkpack/modules/storyteller_dice/code/roll_datum.dm
FalloutFalcon: player mob is the guy we are showing it to. (the mobs around us)

PR #1101, opened 2026-06-03
Scry the Hearthstone bugfixes and enhancements

### LINKED ISSUE #927: Visceratika 2 Scry the Hearthstone bugs
### Client Version:

_No response_

### Issue Summary:

There are two parts to this. One, the roll for it is public despite it being an unnoticeable power which doesn't quite affect other people. And two, it doesn't take into account your guestbook nor does it show true names so you only see descriptors.
Image

I spoke to Chaz and he said that the roll should be private. Additionally, the issue with names can be possibly fixed with a "get guestbook name" macro. 

### Round ID:

_No response_

### Test Merge Information:

_No response_

### Reproduction Steps:

1. Get some people in a room with a gargoyle with visceratika 2
2. Set up guestbook names
3. Cast it
4. See how the names you set don't show and also how people close enough can see your roll

TheCarnalest:
## About The Pull Request
Fixes #927 by making the power use guestbook names instead of mob names as well as a runtime when the power is deactivated and activated within an area. Additionally makes the power give you the direction and distance to players in your area as well as alerting you when people enter and leave it, and adds the TTRPG-accurate restriction that the power can only be used indoors, i.e. within a structure. Also neatens the code up and standardizes one tile of distance to be described as a yard.

dreamseeker_9JHhDzj4Nb

dreamseeker_zIHzLWPp35

dreamseeker_wIsiKb0Yi2

dreamseeker_Kahmmsh43j

## Why It's Good For The Game
Bugfixes, makes the power slightly more useful, and makes it more TTRPG-accurate by not letting Gargoyles use it outside of structures.
## Changelog
:cl:
fix: Scry the Hearthstone now displays guestbook names.
add: Scry the Hearthstone now gives you direction and distance to players in your area as well as alerting you when they leave and enter, but can no longer be used outdoors.
/:cl:


TheCarnalest:
I added comments so that people know these TTRPG mechanics are missing, but in the TTRPG being pinpointed with this power can be resisted with a roll and Obfuscate can be used to evade it. Should I add them or can they be left out?

TheCarnalest:
Also, should I add alerts for when people enter and leave the area you're monitoring? Could be interesting but IDK if it's wanted seeing as they already have thermal vision.

FalloutFalcon:
> I added comments so that people know these TTRPG mechanics are missing, but in the TTRPG being pinpointed with this power can be resisted with a roll and Obfuscate can be used to evade it. Should I add them or can they be left out?

> Also, should I add alerts for when people enter and leave the area you're monitoring? Could be interesting but IDK if it's wanted seeing as they already have thermal vision.

I dont feel 2 strongly about either tbh..

modular_darkpack/modules/powers/code/discipline/visceratika.dm
```diff
@@ -68,45 +59,77 @@
...
+
 /datum/storyteller_roll/scry_the_hearthstone
 	bumper_text = "scry the hearthstone"
 	applicable_stats = list(STAT_PERCEPTION, STAT_AWARENESS)
 	roll_output_type = ROLL_PRIVATE
 
+/datum/discipline_power/visceratika/scry_the_hearthstone/New(datum/discipline/discipline)
+	. = ..()
+
+	scry_roll = new()
```
FalloutFalcon, line 73: kinda intresting. None of our rolls really do this, i dont really see the reason not to lazy load the datums on the off chance we dont need to waste the effort of making them. Though i have considered standerdizing a "main_roll" datum for powers (gifts and discs) so we can parse out some behavoir like displaying the stats used and such in the button tooltip.

modular_darkpack/modules/powers/code/discipline/visceratika.dm
```diff
@@ -178,10 +201,16 @@
 
 	vitae_cost = 0
 
+/datum/discipline_power/visceratika/armor_of_terra/post_gain()
```
FalloutFalcon, line 238: oh huh this never got or gets removed if you loose the disc.

modular_darkpack/modules/powers/code/discipline/visceratika.dm
```diff
@@ -197,12 +226,14 @@
 	duration_length = 1 SCENES // might be too long...
 	cooldown_length = 10 SECONDS
 
-/datum/discipline_power/visceratika/flow_within_the_mountain/try_activate()
-	// placed in try_activate instead of pre_activation_checks so as to not consume blood while running this check
+/datum/discipline_power/visceratika/flow_within_the_mountain/can_activate(atom/target, alert)
+	. = ..()
+	if (!.)
+		return.
```
FalloutFalcon:
```suggestion
		return .
```

modular_darkpack/modules/powers/code/discipline/visceratika.dm
```diff
@@ -68,45 +59,77 @@
...
+	for(var/mob/living/player in (GLOB.player_list - owner))
+		if(get_area(player) != get_area(owner))
+			continue
+
+		var/distance = get_dist(owner, player)
+		var/location_description
+		if (distance == 0)
+			location_description = "close to you"
+		else
+			location_description = "[distance] [distance == 1 ? "yard" : "yards"] [dir2text(get_dir(owner, player))]"
```
FalloutFalcon: oh, if we are gonna display it using "yards" then we should make sure to convert it using the defines.

Image

FalloutFalcon (changes requested):

modular_darkpack/modules/powers/code/discipline/visceratika.dm
TheCarnalest: I will never recover from this

modular_darkpack/modules/powers/code/discipline/visceratika.dm
TheCarnalest: As per the conversation on Discord, one yard is now a tile (and I edited the Corax power that uses yards to have the same range and a disclaimer)

modular_darkpack/modules/powers/code/discipline/visceratika.dm
TheCarnalest: The difference in overhead is basically nonexistent since both optimize different cases (using it multiple times vs. never using it) so I'll just revert it to lazyloading if you'd like

FalloutFalcon (approved):

PR #1107, opened 2026-06-05
Moves disc power effects onto there own layer

FalloutFalcon:
## About The Pull Request
I still think this can be handled better but lets make sure that it doesnt overlap with tg mutations as thats bound to cause issue.
## Why It's Good For The Game
slightly better behavoir, tho i think i can do better.
## Changelog
:cl:
code: discipline power effects are on there own layer
/:cl:


chazzyjazzy (approved):
nice

PR #1109, opened 2026-06-05
Removes `vars["var_name"]` from ritual code

FalloutFalcon:
## About The Pull Request
wtf
## Why It's Good For The Game
Why was it like this, this is bad, just access the var directly.
## Changelog
:cl:
code: ritual book pulls vars from runes in a same method
/:cl:


chazzyjazzy (approved):
wow that is very odd

PR #1110, opened 2026-06-05
Blood research works

FalloutFalcon:
## About The Pull Request
Actually store the information we need (in both living and carbon, these were hella inconsistant for ???)

image

## Why It's Good For The Game
Blood research should acctually work now.
## Changelog
:cl:
fix: Occult blood research should now work
/:cl:


chazzyjazzy (approved):
looks like it will work

PR #1135, opened 2026-06-12
Cleaner behavoir for reusing rolls

FalloutFalcon:
## About The Pull Request
Reworks the stuff for roll cooldowns to just silently return the result if it can rather then, returning out if you failed the roll but making you reroll it if you passed, which is a bit odd.
If it doesnt return the old roll it just generically fails instead. 

Rolls are also now cleared if they pass the roll cool-down, which should cleanup some odd behavior.
## Why It's Good For The Game
better behavoir, should lead to some less weird results
## Changelog
:cl:
refactor: Tweaks how rolls with cool-down use the old results
/:cl:


chazzyjazzy (approved):
looks solid

PR #1151, opened 2026-06-15
Dragging stuff through transfer points come with you

FalloutFalcon:
## About The Pull Request
Makes it itterate things we consider "attached" to moving objects to bring them along as well when people take transfer points.
## Why It's Good For The Game
zmove (stairs) works like this.
## Changelog
:cl:
qol: Going through transfer points brings items your dragging or buckled to with you
/:cl:


FalloutFalcon:
I wanna improve the handling later because rn you loose buckles but this is fine for now.

chazzyjazzy (approved):
that was a smart way to do it

PR #1157, opened 2026-06-15
Allows for the katana to be used in the burning blade ritual via the unused KJ egorium icon state + Small bugfix.

Ambrose997:
## About The Pull Request

Allows the burning blade ritual to be used on the katana, making use of the unused KJ sprite (supposedly there is no intention of it being used on the rebase).

Also fixes a small bug where the original inhand_icon_state isn't saved separately before it's later restored to original_icon_state, could've caused issues if the icon_state and in_hand_icon state were different before the component is applied.
## Why It's Good For The Game

Makes use of a sprite that would otherwise be wasted, opens up more functionality for the burning blade ritual, potentially allowing tremere to ignite the swords of others and not just themselves. Hopefully one day we can get more firey sword sprites so it can be used on any sword.

Also hopefully will let us have a weapon that we can set on fire which can actually be concealed within a bag rather than having a giant-ass unconcealable scythe on our back 24/7. (although upon making this I have discovered the katana is not sheathable, so I might need to look into this via a separate PR...)
## Changelog
:cl:
add: Added the katana as one of the blade options for the Burning Blade ritual.
fix: saves the inhand_icon_state before restoring to original_icon_state to avoid the wrong sprite being used.
/:cl:


modular_darkpack/modules/ritual_thaumaturgy/rituals/burning_blade.dm
```diff
@@ -9,22 +9,23 @@
...
 		return
-
 	if(!ritual_roll_datum)
 		return
-
 	var/charges = ritual_roll_datum.last_sucess_amount
-	scythe.AddComponent(/datum/component/burning_blade, charges)
-	to_chat(last_activator, span_notice("The scythe ignites with an unholy flame for [charges] swings!"))
+	weapon.AddComponent(/datum/component/burning_blade, charges)
+	to_chat(last_activator, span_notice("The [weapon.name] ignites with an unholy flame for [charges] swings!"))
```
FalloutFalcon:
```suggestion
	to_chat(last_activator, span_notice("[weapon] ignites with an unholy flame for [charges] swings!"))
```

modular_darkpack/modules/ritual_thaumaturgy/rituals/burning_blade.dm
```diff
@@ -35,9 +36,15 @@
...
 	weapon.damtype = AGGRAVATED
-	weapon.icon_state = "egorium"
-	weapon.inhand_icon_state = "egorium"
+
+	if(istype(weapon, /obj/item/katana/vamp))
+		weapon.icon_state = "firetana"
+		weapon.inhand_icon_state = "firetana"
+	else
+		weapon.icon_state = "egorium"
+		weapon.inhand_icon_state = "egorium"
```
FalloutFalcon: should should probally just append `_burning` or something onto the end of the icon state instead having to hardcode each type paths icon state in here.

FalloutFalcon (changes requested):

modular_darkpack/modules/ritual_thaumaturgy/rituals/burning_blade.dm
Ambrose997: > should should probally just append `_burning` or something onto the end of the icon state instead having to hardcode each type paths icon state in here.

Yeah I did think about this considering the current lazy way I did it isn't very futureproof. I suppose I could assign the icon states variables within the weapons themselves and then use them in the burning_blade ritual?

```
/obj/item/scythe/vamp
	var/burning_icon_state = "egorium
/obj/item/katana/vamp
	var/burning_icon_state = "firetana"

/datum/component/burning_blade/Initialize(charges)
	(other code here lol)
	weapon.icon_state = weapon.burning_icon_state
	weapon.inhand_icon_state = weapon.burning_icon_state
```


Guessing that's what you meant?

chazzyjazzy: No what Abby means is

say the scythe's sprite is called 'scythe'

thus the burning sprite is called 'scythe_burning'

say the katana's sprite is called 'katana'

thus the burning sprite is called 'katana_burning'

thus when you apply the component, you add _burning at the end of all the icon states, and when returning you remove it

Ambrose997: Ooooooooh I get it.

Ambrose997:
Did what you asked + made it a list that's iterated through rather than spamming weapon vars and locate()'s.

FalloutFalcon (approved):
much better

PR #1167, opened 2026-06-17
Melee Numbers

FalloutFalcon:
## About The Pull Request
Using WTA as a refrence due to it having a much better list of weapons;

image
image

## Why It's Good For The Game
TTRPG accurecy c: 
also attack_difficulty is a cool new concept we get to try out.

## Changelog
:cl:
balance: Melee force and attack difficulty is more inline with TTRPG numbers
/:cl:


chazzyjazzy (approved):
yes the spear ought to be two handed like the base /tg/ spears (and thus a subtype of them as well) - i tried to do that but lacked sprites

PR #1174, opened 2026-06-20
Showing flavor text while masked checks the holder not the reader

FalloutFalcon:
## About The Pull Request
its meant to be a pref for people to have there character pref visible if they always are wearing masks, not for someone ELSE to use it to pierce through everyone elses obscuring.
## Why It's Good For The Game
sigh.
## Changelog
:cl:
fix: Showing flavor text while masked checks the holder not the reader
/:cl:


chazzyjazzy (approved):
wow

PR #1185, opened 2026-06-30
Tweaks to events and city traits

FalloutFalcon:
## Changelog
:cl:
fix: Rainstorms work again
balance: All current events can now only trigger once
balance: if the faulty grid city trait is active it wont ALSO trigger the blackout event. Fuse-boxes should only break once therefore.
balance: Small chance for blackout to only MOSTLY fuck up a fusebox
/:cl:


chazzyjazzy:
why are we lessening events when we have a desperate need to increase the things that need to happen in the game every round

why not either make the priority_announcement work off a prob() or lower the prob() or the damage done to the fuseboxes (or both, or make it contingent on how many fuseboxes currently exist on-map since a disproportionate amnt of them are being broken from the event)

why also make it dependent on 'engineers' - that is just soft-removing the event

also, why make the szlachta attack only have 1 max occurrance, pve is good, it shakes up the round imo

chazzyjazzy (changes requested):
what id like to see is the blackout and the szlachta attack having unchanged max_occurrances, lowering the dmg from the blackout event (which you have already done), lowering the prob() for a fusebox to be damaged from the blackout event, and perhaps making a prob() for the priority_announce to even fire

i would also like to see a removal  of the dependence upon the blackout city trait + engineers being present which all but soft removes the blackout event

FalloutFalcon:
> why are we lessening events when we have a desperate need to increase the things that need to happen in the game every round

My inference right now is that, having something like the blackout happen more then once in a round is not really ADDING much to the round more then a minor headache.
I think very regularly having the same event fire twice and consistently across multiple rounds lessens the impact of these things.

> why not either make the priority_announcement work off a prob() or lower the prob() or the damage done to the fuseboxes (or both, or make it contingent on how many fuseboxes currently exist on-map since a disproportionate amnt of them are being broken from the event)

Im not sure what you mean by this.. The announcement itself isn't the issue but that all the boxes go and blow up a second time.

> why also make it dependent on 'engineers' - that is just soft-removing the event

I might just increase the min_players thing but my primary thought is i want someone acctaully quailfied to be able to respond it it rather then it being a 5 pop round where its not statisticly unlikely that there is zero players with a character that can fix the shit.
In generally the flavor probally should have some sort of proper mechanic or city worker going out to fix these. Say for instance the random shops in the city, without a deticated roll aint no one fixing those. The faction bases? yea force the ghoul to go shock themselves on the camarilla breaker box.

> also, why make the szlachta attack only have 1 max occurrance, pve is good, it shakes up the round imo

See my first part of the response.. We should add more types of pve..

chazzyjazzy (approved):
not super enthused about this but i'll trust your judgment - particularly tying blackouts to rain and reducing szlachta attack (the sewers should be dangerous?)

FalloutFalcon:
> particularly tying blackouts to rain

not sure what u mean

image

modular_darkpack/modules/city_traits/code/negative_traits.dm
```diff
@@ -36,6 +36,9 @@
 	trait_type = STATION_TRAIT_NEGATIVE
 	can_revert = FALSE
 	darkpack_allowed = TRUE
+	trait_to_give = STATION_TRAIT_BLACKOUT
+	newspaper_message = "We continue to recive delays from city officals on estimates when power will be returned city-wide."
```
chazzyjazzy:
```suggestion
	newspaper_message = "We continue to receive delays from city officals on estimates when power will be returned city-wide."
```

chazzyjazzy:
LGTM

PR #1190, opened 2026-07-03
Clan and Tribe keys wont spawn if the map does not have any corresponding doors

FalloutFalcon:
## About The Pull Request
Subsplat keys wont be handed out if NO doors exist on the map that they could unlock. This only includes maploaded doors and wouldn't affect content that spawns POST players joining (hence the addition of a flag that ignores this behavoir)
## Why It's Good For The Game
This wont cover all cases of not wanting someone to have keys, but for most use-cases of just "we dont have a hideout on this map" this should cover most cases and prevent players ahelping "hey where do theses keys go" just for you to have to say nowhere.

Its also alot more programtic and modular then having to edit keys out on the downstream.
And has the bonus on allowing multiple maps to be used that might have different doors on the map.
## Changelog
:cl:
qol: Clan and Tribe keys wont spawn if the map does not have any corresponding doors
/:cl:

### DIFF
--- a/modular_darkpack/modules/cars/code/car.dm
+++ b/modular_darkpack/modules/cars/code/car.dm
@@ -124,6 +124,9 @@
 		access = "[rand(1,9999999)]"
 		AddComponent(/datum/component/door_ownership)
 
+	if(mapload)
+		GLOB.city_door_lock_ids |= access
+
 	// DARKPACK TODO - see about reimplementing this sprite for cars
 	/*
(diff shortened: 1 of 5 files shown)


chazzyjazzy (approved):
this is much better than TFN's version and also smart move checking also for the cars since it couldve been easily overlooked that some cars share some door code accesses rather than generating a new access string when someone walks up to claim the car like an apartment.

PR #1199, opened 2026-07-06
ARMOIRES AND ARMORS - Making Crossbows Useful Again + Leopoldite Plate Armor Buff

SirSwagmeyer:
## About The Pull Request
Fixes the ever-present reload bug on the Crossbow. Also partially-returned the bolt's damage back to what it used to be. 
I've also made the full suits of plate armor in the Leopoldite sanctuary _actually_ protect you from conventional forms of damage. It's plate armor, you should be able to deflect melee blows better than a kevlar vest, right?
## Why It's Good For The Game
Crossbows were relegated to decorative garbage due to the massive projectile nerf, and the reload bug that soon followed. This will remedy that issue and add another weapon into our diversity pool. 
The plate armor in the Leopoldite armory is just... a base SS13 set of plate armor, and it has become a bit of a meme due to how comically low their protection values are relative to the rest of the armor in the game. Note - a _leather trenchcoat_ can deflect bullets better than the plate armor. That's nuts. 
## Changelog
:cl:
balance: buffed the bolt projectile and leopoldite armor a bit.
fix: fixed the crossbow reloading issue

/:cl:

### DIFF
--- a/code/modules/clothing/head/helmet.dm
+++ b/code/modules/clothing/head/helmet.dm
@@ -500,7 +500,7 @@
 	icon_state = "knight_green"
 	inhand_icon_state = "knight_helmet"
 	armor_type = /datum/armor/helmet_knight
-	flags_inv = HIDEMASK|HIDEEARS|HIDEEYES|HIDEFACE|HIDEHAIR|HIDESNOUT
+	flags_inv = HIDEMASK|HIDEEARS|HIDEEYES|HIDEFACE|HIDEHAIR|HIDESNOUT|HIDEFACIALHAIR
 	flags_cover = HEADCOVERSEYES | HEADCOVERSMOUTH
 	resistance_flags = NONE
 	strip_delay = 8 SECONDS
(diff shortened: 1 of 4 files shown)


chazzyjazzy (approved):
balance pr - might wanna do some testing on the damage value of the crossbow bolt, but it's not like we're here to gatekeep. think thats like a two shot against most characters.

SirSwagmeyer:
> balance pr - might wanna do some testing on the damage value of the crossbow bolt, but it's not like we're here to gatekeep. think thats like a two shot against most characters.

Hm, yeah. I tested it out on a few different splats and I agree - gonna knock down the damage significantly, but increase wound damage significantly instead.

Ambrose997:
> balance pr - might wanna do some testing on the damage value of the crossbow bolt, but it's not like we're here to gatekeep. think thats like a two shot against most characters.

Risks not really being useful if the damage isn't high though. You either successfully hit two shots very accurately or the vampire wrecks you lol.

PR #1219, opened 2026-07-14
Fix missing inhand icon state for door shield

### LINKED ISSUE #515: This door doesnt have an inhand icon state
### Client Version:

_No response_

### Issue Summary:

Image

just needs an inhand icon state, probably already exists

### Round ID:

_No response_

### Test Merge Information:

_No response_

### Reproduction Steps:

1. pick up door shield
2. oh god its gone

baabass1:
## About The Pull Request

This pull request fixes the missing in-hand icon state for the door shield item.

The issue was caused by the door shield defining `icon_state = "door"` but not defining `inhand_icon_state`, preventing the correct sprite from being displayed when the item is held.

This change adds the missing

```dm
inhand_icon_state = "door"
```

assignment to the door shield item definition in `modular_darkpack/modules/doors/code/door_item.dm`. No sprite assets or gameplay behavior were modified.

## Why It's Good For The Game

This change restores the intended in-hand appearance of the door shield by ensuring the correct sprite is displayed when the item is held. The fix is minimal, reuses the existing sprite assets, and does not affect gameplay or unrelated systems.

## Changelog

:cl:
fix: Fixed the door shield not displaying its correct in-hand sprite when held.
code: Added the missing `inhand_icon_state` assignment for the door shield item.
/:cl:


FalloutFalcon:
So the issue seems to be specific doors created as a result of bashing a door down. When you bash a door it spawns the item and then copys some data (like icon state) to it. 

Broken type
image

But these are fine for example.
image
image


setting the inhand icon state isnt super necissary, the code by default will use the icon_state instead if none is set.
setting this means that it would stop the working door types from defaulting to the icon_state making them all homogeneous.



This is actually just a case of bad icon assignment, see here the "shop" icon state.
`modular_darkpack/modules/deprecated/icons/lefthand.dmi`/`modular_darkpack/modules/deprecated/icons/righthand.dmi`
image 
That needs to be renamed to match the inventroy icon states
`modular_darkpack/modules/weapons/icons/weapons.dmi`
image

baabass1:
Thank you for the clarification. I renamed the `shop` icon state to `glass_blue` as suggested. The `inhand_icon_state` code change has also been removed.

FalloutFalcon (approved):
Oh thanks, sorry I didnt see your updates because the pr was a bit old.

PR #1223, opened 2026-07-18
Add Discipline keybinds

TheCarnalest:
## About The Pull Request
Adds three new kinds of keybinds to interact with Disciplines: the first for selecting which Discipline to use, the second for quickly activating a specific power in the selected Discipline, and the third being an optional keybind for every individual power. The latter takes up ~40% of the keybind menu when enabled but is still pretty convenient since it only takes one key to activate a power rather than two with the former.

Both default Discipline keybinds and individual power keybinds can be toggled with a config. Keybinds are automatically generated from a define, meaning codebases can set their own maximums for amount and level of Disciplines and have a matching number of keybinds automatically created. 

Things I want to get done before I undraft this PR:
- [x] Beg one of my artist friends to make a sprite for the currently selected Discipline to make it visually distinguishable
- [x] Have a good night's rest and improve the quality of the implementation when I wake up
- [x] Contemplate how this can be made more player-friendly and possibly implement my findings
- [x] Add a config for this feature

Credit to @cheapcents (aka 2Cents on Discord) for the selected Discipline sprite, with some color palette edits made by myself.

## Demonstration

https://github.com/user-attachments/assets/5ebeb92b-fab1-474b-9455-2c398bc22cab

image

## Why It's Good For The Game
Provides a more intuitive and faster alternative to using the mouse to flip between Discipline levels to get the one you want. Also eliminates the possibility of players using macros for Disciplines to gain an advantage by letting all players normally do it as a QOL feature.
 
## Changelog
:cl:
qol: You can now use keybinds to select and cast Disciplines
/:cl:


TheCarnalest:
The keybinds are currently fully functional, anyone can check the PR out and see it for themselves

TheCarnalest:
I erroneously marked this as a feature instead of quality of life at first, pls edit the tags so my good boy points can be preserved

code/__DEFINES/~darkpack/discipline.dm
```diff
@@ -47,3 +47,12 @@
 #define POWER_TYPE_DISCIPLINE "discipline"
 ///Power is a Chi Art of the Kindred of the East game line
 #define POWER_TYPE_CHI_ART "chi_art"
+
+/// The maximum amount of Disciplines a Kindred is capable of having
+#define MAXIMUM_DISCIPLINES 5
+/// The highest level in a Discipline that can be used
+#define MAXIMUM_DISCIPLINE_LEVEL 5
```
FalloutFalcon: bit a misnomer as we already in theory support 6 dots and there is no coded limit to the amount of discs a mob can containt. This is only a keybinding thing atm?

FalloutFalcon (approved):
Otherwise good. Noticeably existing servers will have both of these unbound by default as configs aren't synced like that

code/__DEFINES/~darkpack/discipline.dm
TheCarnalest: yea I originally planned to have these be universal, then realized the codebase enforces no such hard limits (like normal players can get 5 Disciplines while trusted one can have an unlimited amount) and forgot to change it when I pulled the defines out of non-keybind code

TheCarnalest: major update (changing those define names) incoming

TheCarnalest:
> Otherwise good. Noticeably existing servers will have both of these unbound by default as configs aren't synced like that

My hope is that they see "Config Update" and take a look, or I'll just directly pester people like chaz and linnepap to update their configs

PR #1238, opened 2026-07-23
Fixes guestbook revealing known names even while masked

Hardly3D:
## About The Pull Request
Adds a check on whether our target is wearing a mask, keeping them unknown if they're masked up or have opted in to show their identity through character prefs.
Renamed ``show_flavor_text_when_masked`` to ``show_identity_when_masked`` to reflect this change as well.

https://github.com/user-attachments/assets/02a8b025-178b-4c17-b824-94cdde18d6f2

## Why It's Good For The Game
Lets people actually be anonymous while wearing a mask and not have to worry about getting metagamed.

image

## Changelog
:cl: Hardly
fix: Fixes guestbook revealing names even on masked people
/:cl:


modular_darkpack/modules/guestbook/code/guestbook.dm
```diff
@@ -136,8 +136,9 @@
 	if(guest)
 		if(user == guest)
 			return guest.real_name
-		else
-			checked_name = guest.real_name
+		if(guest.get_face_name() == "Unknown")
+			return NONE
```
FalloutFalcon: NONE implies its a bitflag.
```suggestion
			return null
```

FalloutFalcon (changes requested):
This should have a toggle or respect `/datum/preference/toggle/show_flavor_text_when_masked` (though you might want to rename it to reflect it being more generic.)
There is a meaningful amount of players that have a character that permanently wears a mask so having them be permanently unknown would be frustrating.

FalloutFalcon (approved):

PR #1247, opened 2026-07-25
Dropdown for character slots in preferences

FalloutFalcon:
## About The Pull Request
Modified version of part of https://github.com/NocturneWoD/NocturneWoD-SecondCity/pull/17 which i think is just from another hrp server but it doesnt say where and i've definitly seen it in like 5 other places.
## Why It's Good For The Game
A billion buttons looks bad.
## Changelog
:cl:
qol: If you have more then 5 character profiles in preferences, it displays as a dropdown instead.
/:cl:


TheCarnalest (approved):
JavaScript isn't a strength of mine, but it works well and looks good so this gets my approval

PR #1258, opened 2026-07-26
Acid and blood can be scooped from the respective tiles

FalloutFalcon:
## About The Pull Request
scoop my goop
## Why It's Good For The Game
That shit is not water.
Having an infinite source of blood is probs weird but i would need to rework the reagent code to be a list of percent extracted and i cant be asked.
## Changelog
:cl:
fix: Non-water turfs now have the correct scoop-able material
/:cl:

### DIFF
--- a/modular_darkpack/modules/walls/code/floors/water.dm
+++ b/modular_darkpack/modules/walls/code/floors/water.dm
@@ -27,6 +27,7 @@
 	light_range = 1
 	light_power = 0.5
 	baseturfs = /turf/open/water/acid
+	reagent_to_extract = /datum/reagent/toxin/acid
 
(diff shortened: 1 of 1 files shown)


TheCarnalest (approved):
scoop my goop twin

TheCarnalest (commented):
it's not a problem since you'd expect to see sewage made of mostly water, but do we have anything that'd better represent `/turf/open/water/vamp_sewer` actually?

FalloutFalcon:
> it's not a problem since you'd expect to see sewage made of mostly water, but do we have anything that'd better represent `/turf/open/water/vamp_sewer` actually?

I would prob want to mix it with a minor toxin but still mostly be water reagent. Tho the scoop-able element doesnt support that. I made the element in tg to standardize the plasma scooping interaction but didn't consider multiple reagents.

PR #1266, opened 2026-07-26
Hunger tweaks

FalloutFalcon:
## About The Pull Request
Moves hunger multipliers from https://github.com/DarkPack13/SecondCity/pull/857 to a config rather then being in two seperate places which made it so our code has a * 0.04 of normal tg which is why no one got hungry.
Default config value is set to 0.5 of normal TG values. This is something that should likely be tweaked.

## Why It's Good For The Game
hunger rate is pretty subjective and depends on the length of the round and population imo. 

Having a ton of stores/food/cooking but NO reason to ever eat is bad. I might follow this up with a minor buff for eating player made food.

## Changelog

:cl:
fix: Hunger is no longer accidentally soft-removed
config: Hunger rate is on a config for easier tweaking
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/physiology.dm
+++ b/code/modules/mob/living/carbon/human/physiology.dm
@@ -41,7 +41,7 @@
 
 	var/datum/armor/armor // internal armor datum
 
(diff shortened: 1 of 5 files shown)


chazzyjazzy (approved):
smart

PR #1271, opened 2026-07-27
Highest ranking wolfs can adjust renown of lower ranks

FalloutFalcon:
## About The Pull Request
Adds a mechanic to adjust renown that is not ahelping.

Adds an action for high ranking wolfs (councillor right now as im unsure how i would do this for bsd rn.) to be able to adjust any of the 3 values of any werewolf or kinfolk (since its technicly something they have in code).  This is logged and they have to provide a reason so administrative issues should not be too bad.

I foresee a similar addition for masquerade as I think it suffers from alot the same issues for how automatic is handled

In future I might add a config for this where you can opt into a automatic system more close to the masq where its all automatic. (and vice versa for the masq) I'm really not sold the best approach and i think it varies a bit based on servers.
Its not TOO different from the fact that a player can report your massq breach however.
## Why It's Good For The Game
Renown fufills a similar niche as masq were its a mutable value that is primarily a role-play mechanic which shifts across rounds and while persistent across rounds is still a temporary thing that you can loose or gain quite easily. 
It (and dignitas) are both functionally reputation meters whos realistic dishes are all currently fulfilled by player characters.
## Changelog
:cl:
add: Leaders of the Garou Nation can change the renown of other wolves.
/:cl:

### DIFF
--- a/code/__DEFINES/~darkpack/fera/werewolf_renown.dm
+++ b/code/__DEFINES/~darkpack/fera/werewolf_renown.dm
@@ -11,3 +11,5 @@
 #define RANK_ATHRO 4
 #define RANK_ELDER 5
 #define RANK_LEGEND 6
+
+#define MAX_RENOWN 10
--- a/modular_darkpack/modules/jobs/code/garou/councillor.dm
+++ b/modular_darkpack/modules/jobs/code/garou/councillor.dm
@@ -31,6 +31,11 @@
 		"Guardian"
 	)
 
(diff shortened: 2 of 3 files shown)


FalloutFalcon:
Im actually gonna close this. I realize that **10** temp renown converts into 1 permanent renown (which is all we really represent rn) All the charts use temp willpower for how they suggest awards. Which means in order to engage with this mechanic faithfully atm someone would have to internally keep track of there temp renown. This kinda blows.
This is gonna need a rework to this interaction first.. #1089 probs.

PR #1310, opened 2026-07-31
Attribute/Abilities description overhaul.

Major00:
## About The Pull Request

I've altered the descriptions to better suit our current direction. It now simply explains what the abilities, virtues, morality, and abilities do in a role playing sense.

Also I fixed some strange wording and made what they represent more clear such as science being about the physical sciences and academics being about the social sciences.

## Why It's Good For The Game

It makes new players less confused on apocrypha.

## Changelog

:cl:

fix: altered the descriptions to be less vampire focused/ mechanics focused. 

spellcheck: fixed some grammatical errors within the descriptions.

/:cl:

### DIFF
--- a/modular_darkpack/modules/storyteller_stats/code/st_stats/default_abilities.dm
+++ b/modular_darkpack/modules/storyteller_stats/code/st_stats/default_abilities.dm
@@ -2,7 +2,7 @@
 /datum/st_stat/ability/alertness
 	subcategory = "Talents"
 	name = "Alertness"
-	description = "Affects your character's intuition."
+	description = "Affects your character's vigilance, alongside their ability to understand and react to threats."
 
 /datum/st_stat/ability/athletics
 	subcategory = "Talents"
@@ -12,7 +12,7 @@
 /datum/st_stat/ability/awareness
 	subcategory = "Talents"
 	name = "Awareness"
-	description = "Affects your character's attention to extreme detail. Used in Auspex."
+	description = "Affects your character's ability to notice otherworldly and supernatural events for what they are."
 
 /datum/st_stat/ability/brawl
 	subcategory = "Talents"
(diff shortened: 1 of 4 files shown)


modular_darkpack/modules/storyteller_stats/code/st_stats/default_abilities.dm
```diff
@@ -32,12 +32,12 @@
...
 	subcategory = "Talents"
 	name = "Intimidation"
-	description = "Affects your character's ability to persuade via coercion and violence. Used in Dominate, Presence and other mental and social disciplines."
+	description = "Affects your character's ability to persuade via coercion and violence."
 
 /datum/st_stat/ability/leadership
 	subcategory = "Talents"
 	name = "Leadership"
-	description = "Affects your character's natural affinity to lead, direct, take responsibility, and strategize. Increases your maximum follower/minion limit."
+	description = "Affects your character's ability to lead, direct, and organize groups."
```
chazzyjazzy, line 40: This is just a direct downgrade - instead of now clearly communicating what the leadership stat does, "Increases your maximum follower/minion limit", it instead makes it a bit more unclear. Consider that we need to clearly communicate what these stats are used for to new players - instead of having to communicate them through a third party source like a wiki or an mhelp or discord, or, oh god, ooc chat

chazzyjazzy (changes requested):

Faengi:
A lot of this seems like a step back in documentation? Instead of just telling players what they're getting, this seems to make a lot of it vague-r, which I don't really see the advantage of; it just means new players will have to ask in OOC chat or Discord or code dive.

FalloutFalcon:
I think most of it is fine, I think specifically ditching discipline info as its quite out of date and doesn't display gifts, while probably being better suited as a codified thing since. Tho chaz is right for the leadership atleast that we can describe the mechanics in a bit more depth still as it helps players make choices.

Major00:
If the core problem is that tfn and monke want it to be more mechanics focused descriptions i'd be happy to utilize some kind of toggle so the description could change based on which server you're using it for. Or even just pring something like this to Apocrypha.

chazzyjazzy:
> If the core problem is that tfn and monke want it to be more mechanics focused descriptions i'd be happy to utilize some kind of toggle so the description could change based on which server you're using it for. Or even just pring something like this to Apocrypha.

i think it goes a bit deeper than mechanics vs roleplay focused descriptions. yeah we should try to disincentivize players from powergaming, and specifically measuring out in the tooltip what exact mechanical effect a stat has can certainly assist in that task. but perhaps obfuscating it on purpose is... 'hostile' game design? idk. i just know that a common complaint is 'wtf does this stat do' and the first point of contact for a complaint like that is 'read the tooltip' and this makes that solution less helpful.

up to you what youd like to do

>tfn

tfn no longer exists and shut down last night

PR #1334, opened 2026-08-03
st_stat changes that affects the mob is handled better

FalloutFalcon:
## About The Pull Request
Shifts them from flags it checks for in a loop to just directly calling a proc as only one stat has these flags making them kinda pointless.
## Why It's Good For The Game
Ensures its easier to write code that changes mob stats based on there stats and that its easy to make sure its ALWAYS updated and doenst need a helper called after every update.
## Changelog
:cl:
code: Changing stats based on stamina or dexterity is handled smarter
/:cl:


chazzyjazzy (approved):
oh i see, this is alot better

PR #1340, opened 2026-08-05
Nerfs Acute Hearing

Stutternov:
## About The Pull Request

I've been using this trait and it feels a bit.. too strong.

A person standing 4 tiles away from me can be heard with full clarity, as the 'hearing distance' is effectively the number assigned +1

With this change it changes it from clear coherent hearing at 4 tiles apart down to 2. So you can hear whispers clearly 1 extra tile away, and jarbled ones at another tile after that.

This change does nerf it but it feels still useful, given even if you can't hear full whispers ~3 tiles away from you, you get some of the jarbled text to match it up.

## Why It's Good For The Game

This made whispering really pointless with how many people can hear it. Sure, there's a level of masquerade to it but it makes entire non-masq breaching things whispered heard by a lot of people due to how cheap the trait is.

I feel either this should be merged, or a increase in the merit's cost if it has that range.
 
## Changelog

:cl:
balance: Reduces bonus tiles on acute hearing from 3 to 1 (hearing whispers 4 tiles away down to 2)
/:cl:


FalloutFalcon (approved):
Its a 1 point merit. This is kinda wild. Fits alot closer with some of the other traits it grants.

chazzyjazzy (approved):

PR #1346, opened 2026-08-06
don't spawn NPCs during unit tests (+minor `SShumannpcpool` fixes)

Absolucy:
## About The Pull Request

this simply disables the NPC spawning pool stuff during unit tests. because spawning random NPCs and respawning them if they die is _asking_ for flaky failures.

also properly remove npcs from from `SShumannpcpool.currentrun` when they're destroyed and slightly cleans up `SShumannpcpool.fire()`

## Why It's Good For The Game

less overhead and issues for unit tests

## Changelog

no player-facing changes at all, only affects unit tests

### DIFF
--- a/modular_darkpack/modules/npc/code/human/__npc.dm
+++ b/modular_darkpack/modules/npc/code/human/__npc.dm
@@ -133,6 +133,7 @@
 	drop_on_death_list = null
 	GLOB.npc_list -= src
 	GLOB.alive_npc_list -= src
+	SShumannpcpool.currentrun -= src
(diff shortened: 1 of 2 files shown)


FalloutFalcon (approved):
They are still create and destroyed, if we want a test specifically to test ai, something like monkey business would prob be more suitible. There whole code is shit from the butt though and thats for after there ai rework.

Absolucy:
>They are still create and destroyed

yeah i know, this just gets rid of the bajillion extra npcs that aren't even part of the actual unit testing

PR #1348, opened 2026-08-06
Minor Gun and Cargo Tweaks

breadcrombs:
## About The Pull Request
This pr brings the lever action rifle's storage in line with similar weapons (hunting rifle, carbine, aug) by making it storable in the belt slot and not just the back slot.

Also makes base .45 ACP ammo buyable from the warehouse which somehow only existed in npc gunshops.
## Why It's Good For The Game
The lever action rifle was weird being the only rifle that can't be stored in the belt while similar rifles such as the hunting rifle could, this made the lever action rifle feel weird and inconsistent. 

Cargo not being able to order to order base .45 ACP but a subtype of .45 ACP silver was also a bit weird. 

Making the lever action and db cargo buyable is also good if the hunting shop npc somehow dies as a alternative way to aquire said guns similar to the hunting rifle.
## Changelog
:cl:

add: The lever action rifle alongside the double barrel shotgun can now be purchased from the supply console.
add: 45 ACP can now be bought from the supply console.
balance: The lever action rifle can now be stored in the belt slot.
/:cl:


chazzyjazzy (approved):

Marshmellow105:
HV and HP are meant to be FBI exclusive to give them a leg up ingame, I dont think giving it to everyone is a good idea considering it'll just mean people dont buy normal 9 or 45 anymore because the direct upgrade is available at a practically free pricepoint

breadcrombs:
> HV and HP are meant to be FBI exclusive to give them a leg up ingame, I dont think giving it to everyone is a good idea considering it'll just mean people dont buy normal 9 or 45 anymore because the direct upgrade is available at a practically free pricepoint

HP is more of a sidegrade due to no AP and HV is more of a small upgrade, also this gives the option of the FBI to restock their HP rounds, I do agree with the somewhat free pricepoint for HV so I will increase the price to make it specialized for those who only can carry a small arm due to one reason or another.

kimesik:
> HP is more of a sidegrade due to no AP

FMJ .45 ACP has 20 damage and 5 AP.
HP .45 ACP has 25 damage and 0 AP.
In practice there's no difference in damage against armoured targets between FMJ and HP.
Meanwhile, HP has +5 wound bonus, which is pretty brutal for a pistol.
HP is a straight upgrade.

> also this gives the option of the FBI to restock their HP rounds

As a FBI agent I've never ran out of HP ammo. The starting box has quite a bit of ammo to begin with, and if you do run out, it might be a sign that it's time to switch to bigger guns.

PR #1356, opened 2026-08-07
Dollar Stack Size Adjustment

Stutternov:
## About The Pull Request

Very simple 1 line change - sets the 'full' (1,000 ) dollar stack to be small size instead of deciding to up itself to a normal sized item suddenly.

Dollars are tiny, so a full stack should be a small wad of cash. This makes 1,000 dollar stacks able to fit in boxes and such but they'll still take up space.

## Why It's Good For The Game

Money was very cumbersome where 8k was enough to fully fill a bag. Now it would take a bit more than 8k, making money still cumbersome due to the fact it's a small item - not tiny - so it'll fill inventory slots up still. But now it'll fit in your pocket, your wallet, a box, etc.

## Changelog
:cl:
add: Adjusts size of 1,000 dollar stack from normal to small.
/:cl:

### DIFF
--- a/modular_darkpack/modules/economy/code/dollar.dm
+++ b/modular_darkpack/modules/economy/code/dollar.dm
@@ -8,6 +8,7 @@
 	righthand_file = null
 	ONFLOOR_ICON_HELPER('modular_darkpack/modules/deprecated/icons/onfloor.dmi')
 	w_class = WEIGHT_CLASS_TINY
+	full_w_class = WEIGHT_CLASS_SMALL
 	resistance_flags = FLAMMABLE
 	max_amount = 1000
 	merge_type = /obj/item/stack/dollar


FalloutFalcon (approved):

FalloutFalcon (approved):
~~Ah yea ok had to reread because I saw spaccash at tiny and assumed i made a mistake-~~

Alternatively you could increase the max_amount to something that makes sense for WEIGHT_CLASS_NORMAL which would make it easier to deal in large amounts.

Stutternov:
> Alternatively you could increase the max_amount to something that makes sense for WEIGHT_CLASS_NORMAL which would make it easier to deal in large amounts.

I was going to do that initially but I realized the issue with that is if we make it - say, 5,000 dollars, and I put 4,999 in a stack - it would still be tiny item instead of normal size until I fill it.

I think it's probably fine to do it this way instead just to avoid silly cases like that, plus I do like cash being a bit cumbersome to encourage taking some level of risk in depositing it in ATMs to avoid the clutter.

chazzyjazzy (approved):

Iamgoofball:
You should be able to stack 5000. The problem with ATMs right now is they're way too easy to destroy so nobody can actually use ATMs as ATMs.

